PR #13, opened 2020-12-08
Add QR-DQN

### LINKED ISSUE #12: Implement QR-DQN
Paper: https://arxiv.org/abs/1710.10044

@ku2482 Please tell us if you want to work on that ;)

ku2482:
Implement QR-DQN

## Description

Paper: https://arxiv.org/abs/1710.10044
- [x] Implement QR-DQN
- [x] Add quantile_huber_loss to `sb3_contrib.common.utils`
- [x] Add document for QR-DQN
- [x] Add benchmark

## Context

closes #12 

## Types of changes

## Checklist:

- [x] I have updated the changelog accordingly (**required**).

### Benchmark

|Environments|   QRDQN   | DQN
|------------|-----------| -----------|
|            |logs_qrdqn/| |
|Breakout    |413 +/- 21 | ~300 |
|Pong        |20 +/- 0   | ~20 |

|Environments|    DQN    |   QRDQN   |
|------------|-----------|-----------|
|            |logs_qrdqn/|logs_qrdqn/|
|CartPole    |386 +/- 64 |500 +/- 0  |
|MountainCar |-111 +/- 4 |-107 +/- 4 |
|LunarLander |168 +/- 39| 195 +/- 28|
|Acrobot     |-73 +/- 2  |-74 +/- 2  |

Results_Breakout
Results_Pong

Results_Acrobot
Results_CartPole
Results_LunarLander
Results_MountainCar


sb3_contrib/__init__.py
```diff
@@ -1,6 +1,7 @@
 import os
 
 # from sb3_contrib.cmaes import CMAES
```
araffin, line 3: minor: could you remove that commented line? (my fault)

sb3_contrib/qrdqn/policies.py
```diff
@@ -0,0 +1,244 @@
...
+        action_space: gym.spaces.Space,
+        lr_schedule: Schedule,
+        n_quantiles: int = 200,
+        net_arch: Optional[List[int]] = None,
+        activation_fn: Type[nn.Module] = nn.ReLU,
+        features_extractor_class: Type[BaseFeaturesExtractor] = FlattenExtractor,
+        features_extractor_kwargs: Optional[Dict[str, Any]] = None,
+        normalize_images: bool = True,
+        optimizer_class: Type[th.optim.Optimizer] = th.optim.Adam,
+        optimizer_kwargs: Dict[str, Any] = {"eps": 0.01 / 32},
```
araffin, line 117: Please avoid mutable as default argument
why `0.01 / 32`, it depends on the minibatch size?

See https://github.com/satwikkansal/wtfpython#-beware-of-default-mutable-arguments

sb3_contrib/common/utils.py
```diff
@@ -0,0 +1,30 @@
...
+    Partially taken from https://github.com/bayesgroup/tqc_pytorch
+
+    :param current_quantile: current estimate of quantile value
+    :param target_quantile: target quantile value
+    :return: the loss
+    """
+    n_quantiles = current_quantile.shape[-1]
+    tau = (th.arange(n_quantiles, device=current_quantile.device).float() + 0.5) / n_quantiles
+
+    if current_quantile.ndim == 3:
```
araffin, line 16: TQC has more dimensions because there are several critics?

araffin (commented):
Overall, look good already =)

Obviously missing tests, documentation and benchmark but it is a good start!

Did you had any issue so far? or things that slow you down because of how SB3 work in implementing QR-DQN?
(asking for what we can improve in SB3 ;))

sb3_contrib/qrdqn/policies.py
ku2482: >why 0.01 / 32, it depends on the minibatch size?

It's proposed in the paper (they swept it in  (0.01/32, 0.005/32, 0.001/32)), and 32 is the minibatch size.

araffin: ok, probably put a comment too then ;)

ku2482:
@araffin 
Hi, I have implemented QR-DQN roughly, and I have some points I want to discuss.

1
I named some classes and variables like `QuantileNetwork` and `current_quantile`, however in TQC there are some variables like `next_z`. Could you tell me what names do you prefer?

2
I added `quantile_huber_loss` to `sb3_contrib.common.utils`. However, if you plan to add IQN([paper](https://arxiv.org/abs/1806.06923)) later, `tau` should be an argument. What do you think?

3
I didn't inherit any classes from DQN's for simplicity. I'm familiar with RL, but not with software architecture, and I'm not sure if I should inherit and re-use it. Could you give me any advice?

4
I don't have enough computational resources to benchmark on Atari.
Could you help me test the performance once I finish implementing it?

(I only have 16GB of RAM and I usually store images as a list of LazyFrames, which uses 4 times less memory. Maybe it's better to add such a feature to reduce memory usage?)

Thanks ;)

sb3_contrib/common/utils.py
ku2482: Yes.

araffin:
>, however in TQC there are some variables like next_z. Could you tell me what names do you prefer?

`current_quantile` is must better ;) (please change TQC too so the two matches, probably `next_quantiles` if it is `next_z`)

> I added quantile_huber_loss to sb3_contrib.common.utils. However, if you plan to add IQN(paper) later, tau should be an argument. What do you think?

what is tau exactly? (it would be nice if we have a better name, I'm quite new to quantile regression, so everything is not too clear yet)
yes, it sounds good if it does not add too much complexity.
(I need to re-read IQN later)

>I didn't inherit any classes from DQN's for simplicity.

that's fine ;)
It is part of the design choices we made for SB3.
We favor simplicity and readability over modularity. (even though we try to avoid code duplication when possible)

> Could you help me test the performance once I finish implementing it?

yes ;) but please test it first on simpler env and with a smaller replay buffer size.

>I only have 16GB of RAM and I usually store images as a list of LazyFrames, which uses 4 times less memory. Maybe it's better to add such a feature to reduce memory usage?

Look at the zoo and the replay buffer, we have a `optimize_memory_usage` argument for that ;)

sb3_contrib/common/utils.py
araffin: then probably add a small comment next to the `For TQC`

ku2482:
>please change TQC too so the two matches, probably next_quantiles if it is next_z

>We favor simplicity and readability over modularity. (even though we try to avoid code duplication when possible)

> yes ;) but please test it first on simpler env and with a smaller replay buffer size.

Sure, thank you for your advice.

>what is tau exactly?

We model the quantile function, which is the mapping from the cumulative probability to the quantile value.
`tau` is the cumulative probabilities, which are fixed at equal intervals in TQC and QR-DQN (not in IQN).
So, `cum_p` or `cum_prob` would be a good candidate?

>Look at the zoo and the replay buffer, we have a optimize_memory_usage argument for that ;)

I didn't notice it. It seems it stores the next observation and observation as the same array, and so efficient 👍 

>  I'm not sure for the lazy frame, but if it is simple enough to implement,

[LazyFrames](https://github.com/Stable-Baselines-Team/stable-baselines/blob/master/stable_baselines/common/atari_wrappers.py#L244) is a list of arrays, and convert a list into a (frame-stacked) array when called. Because we store it as a list of arrays, which exactly is a list of references to arrays, it never store the same frame twice.
It's really simple to implement as [this](https://github.com/ku2482/sac-discrete.pytorch/blob/master/sacd/memory/base.py#L46), but it is a little bit slow.


Thank you for your kind response.

araffin:
> tau is the cumulative probabilities, which are fixed at equal intervals in TQC and QR-DQN (not in IQN).

thanks for the refresher =).
So, if I remember correctly, this is the case for QR-DQN because each quantile is a "unit quantile" (in the sense each of them have the same weights)?

>cum_prob would be a good candidate?

yep + a comment ;) (in case of doubt, more verbose name is always good ;))

>I didn't notice it. It seems it stores the next observation and observation as the same array, and so efficient +1

yes, it is not as memory efficient as LazyFrame but it is as fast as a normal buffer and works also without frame stacking at the cost of some complexity (that's why it is False by default).

>It's really simple to implement as this, but it is a little bit slow.

how much slower? (1.2x or ~2x slower)

sb3_contrib/qrdqn/policies.py
ku2482: What is the best way to implement default "eps"?

araffin: >What is the best way to implement default "eps"?


there are different ways...
we do it like that in A2C: https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/a2c/a2c.py#L111

ku2482:
> So, if I remember correctly, this is the case for QR-DQN because each quantile is a "unit quantile" (in the sense each of them have the same weights)?

Yes 👍 

> how much slower? (1.2x or ~2x slower)

I will test it again in a couple of days.

sb3_contrib/qrdqn/policies.py
ku2482: Thank you :)

sb3_contrib/qrdqn/__init__.py
```diff
@@ -0,0 +1,2 @@
+from sb3_contrib.qrdqn.policies import CnnPolicy, MlpPolicy
```
araffin, line 1: you will need to update `setup.cfg` to ignore the warnings of the linter ;)

sb3_contrib/qrdqn/policies.py
```diff
@@ -0,0 +1,248 @@
...
+        self.net_args = {
+            "observation_space": self.observation_space,
+            "action_space": self.action_space,
+            "n_quantiles": self.n_quantiles,
+            "net_arch": self.net_arch,
+            "activation_fn": self.activation_fn,
+            "normalize_images": normalize_images,
+        }
+
+        if optimizer_class is th.optim.Adam and "eps" not in self.optimizer_kwargs:
```
araffin, line 149: I would rather do it in the contructor of `QRDQN` so you also have access to the batch size there ;)

sb3_contrib/qrdqn/qrdqn.py
```diff
@@ -0,0 +1,244 @@
...
+
+from sb3_contrib.common.utils import quantile_huber_loss
+from sb3_contrib.qrdqn.policies import QRDQNPolicy
+
+
+class QRDQN(OffPolicyAlgorithm):
+    """
+    Quantile Regression Deep Q-Network (QR-DQN)
+    Paper: https://arxiv.org/abs/1710.10044
+    Default hyperparameters are taken from the paper
```
araffin, line 19:
```suggestion
    Default hyperparameters are taken from the paper
    
```

sb3_contrib/qrdqn/qrdqn.py
```diff
@@ -0,0 +1,244 @@
...
+            # Sample replay buffer
+            replay_data = self.replay_buffer.sample(batch_size, env=self._vec_normalize_env)
+
+            with th.no_grad():
+                # Compute the target quantile values
+                target_quantile = self.quantile_net_target(replay_data.next_observations)
+                # Follow greedy policy: use the one with the highest value
+                target_quantile, _ = target_quantile.max(dim=2)
+                # 1-step TD target
+                target_quantile = replay_data.rewards + (1 - replay_data.dones) * self.gamma * target_quantile
```
araffin, line 161: probably rename it `quantile_backup` so you don't overwrite `target_quantile` defined above?
btw, why `target_quantile` (singular) and not `target_quantiles` ?

sb3_contrib/tqc/tqc.py
```diff
@@ -242,19 +223,18 @@ def train(self, gradient_steps: int, batch_size: int = 64) -> None:
                 next_actions, next_log_prob = self.actor.action_log_prob(replay_data.next_observations)
                 # Compute and cut quantiles at the next state
                 # batch x nets x quantiles
-                next_z = self.critic_target(replay_data.next_observations, next_actions)
-                sorted_z, _ = th.sort(next_z.reshape(batch_size, -1))
-                sorted_z_part = sorted_z[:, : self.critic.quantiles_total - top_quantiles_to_drop]
+                next_quantile = self.critic_target(replay_data.next_observations, next_actions)
```
araffin, line 226: dito (about singular/plural)

sb3_contrib/qrdqn/qrdqn.py
araffin: this is for the automatic documentation

sb3_contrib/tqc/tqc.py
ku2482: >probably rename it quantile_backup so you don't overwrite target_quantile defined above?

(Belleman) "buckup" and "target" mean the same thing here, don't they?
For clarity, it may be good to name them `next_*` and `target_*`? (it would be incompatible with stable_baselines3)
What do you think?
```
next_* = ...  # Estimate values for next observation
target_* = ... next_* ...   # Something like "R + gamma * next_*", which is the target for the loss
```

>btw, why target_quantile (singular) and not target_quantiles ?

I was confused that some variables like `next_log_prob`, `current_q`, and `target_q` are singular in DQN and SAC, but others are plural. There was no particular intention, so I'd be happy to rename variables as you prefer :)

araffin: >(Belleman) "buckup" and "target" mean the same thing here, don't they?

yes, I was using "target" in SAC/TD3 because the next q-value came from the target network.
but I would agree with renaming `target_` to `next_` (and maybe rename the first `next_target_quantiles`), what do you think?

(we would need to do that in https://github.com/DLR-RM/stable-baselines3/pull/245 to keep SB3 in sync)

>current_q, and target_q are singular in DQN and SAC, 

sorry for that, I guess the true name should be `current_q_estimates` (plural), maybe a thing to change in https://github.com/DLR-RM/stable-baselines3/pull/245 (by @erniejunior )
and `log_prob` is short for "log probabilities" (I truncated it to avoid too long names and equations on several lines) but I agree it is a bit confusing.

ku2482: > I was using "target" in SAC/TD3 because the next q-value came from the target network.
> but I would agree with renaming target_ to next_ (and maybe rename the first next_target_quantiles), what do you think?

I see, now it makes sense. But I assume people call it "target" because it's the target value for training, not because it comes from the target network.
In my opinion, because "buckup" is a less common word, it's clearer for me either
1. to name both "Q(s_t+1)" and "R + gamma * (max)Q(s_t+1)"  as `target_*`, as in [DQN's code](https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/dqn/dqn.py#L157)
2. to name "Q(s_t+1)" as `next_*` and "R + gamma * (max)Q(s_t+1)" as `target_*`, as in (my) [TQC's code](https://github.com/ku2482/stable-baselines3-contrib/blob/feat/qrdqn/sb3_contrib/tqc/tqc.py#L220).

> sorry for that, I guess the true name should be current_q_estimates (plural)

I see. So it's natural to name `current_quantile` or `current_quantile_estimates`?

araffin: I would go for 2. (and we will need to change that in https://github.com/DLR-RM/stable-baselines3/pull/245.)

>I see. So it's natural to name current_quantile or current_quantile_estimates?

here it is slightly different as you have several quantiles (the complete name of `current_q_estimates` would be `current_q_values_estimates` maybe `current_q_values` would be better in fact)

ku2482: >here it is slightly different as you have several quantiles (the complete name of current_q_estimates would be current_q_values_estimates maybe current_q_values would be better in fact)

I see. I'm not fluent in English, so I'll follow your preference.

ku2482: I pushed some commits to rename them, what do you think?
(I modified TQC's code somewhat to make it easy to understand, sorry about that...)

Also, how should I evaluate QR-DQN?
The evaluation protocol of sb3 seems different from the paper.
(The paper reported the best agent's score within 125k steps(500k frames))

Anyway, I really appreciate your kind support ;)

sb3_contrib/common/utils.py
```diff
@@ -0,0 +1,36 @@
...
+        )
+    if current_quantiles.ndim not in (2, 3):
+        raise ValueError(f"Error: The dimension of current_quantiles ({current_quantiles.ndim}) needs to be either 2 or 3.")
+
+    n_quantiles = current_quantiles.shape[-1]
+
+    # Cumulative probabilities to calculate quantile values.
+    cum_prob = (th.arange(n_quantiles, device=current_quantiles.device, dtype=th.float) + 0.5) / n_quantiles
+    if current_quantiles.ndim == 3:
+        cum_prob = cum_prob.view(1, 1, -1, 1)  # For TQC
```
araffin, line 28: It would be nice to have the meaning of each dimension in comment (as it was done before)
and add the comment on why the shape for TQC is different that for QR-DQN

sb3_contrib/qrdqn/qrdqn.py
```diff
@@ -0,0 +1,250 @@
...
+
+            # Get current quantile estimates
+            current_quantiles = self.quantile_net(replay_data.observations)
+
+            # Retrieve the quantiles for the actions from the replay buffer
+            actions = replay_data.actions[..., None].long().expand(batch_size, self.n_quantiles, 1)
+            current_quantiles = th.gather(current_quantiles, dim=2, index=actions).squeeze(2)
+
+            # Compute Quantile Huber loss
+            loss = quantile_huber_loss(current_quantiles, target_quantiles) * self.n_quantiles
```
araffin, line 176: why do you multiply by `self.n_quantiles`? (that was not done in TQC original code if I recall... I did not check for QR-DQN yet)

sb3_contrib/qrdqn/qrdqn.py
```diff
@@ -0,0 +1,250 @@
...
+                # Follow greedy policy: use the one with the highest value
+                next_quantiles, _ = next_quantiles.max(dim=2)
+                # 1-step TD target
+                target_quantiles = replay_data.rewards + (1 - replay_data.dones) * self.gamma * next_quantiles
+
+            # Get current quantile estimates
+            current_quantiles = self.quantile_net(replay_data.observations)
+
+            # Retrieve the quantiles for the actions from the replay buffer
+            actions = replay_data.actions[..., None].long().expand(batch_size, self.n_quantiles, 1)
```
araffin, line 172: please put a comment about what shape is actions after that transform

sb3_contrib/qrdqn/qrdqn.py
```diff
@@ -0,0 +1,250 @@
...
+            current_quantiles = th.gather(current_quantiles, dim=2, index=actions).squeeze(2)
```
araffin, line 173:
```suggestion
            current_quantiles = th.gather(current_quantiles, dim=2, index=actions).squeeze(dim=2)
```

Please use explicit keywords when dealing with dimensions ;)

sb3_contrib/tqc/tqc.py
```diff
@@ -237,24 +218,27 @@ def train(self, gradient_steps: int, batch_size: int = 64) -> None:
...
 
-            # Get current Q estimates
-            # using action from the replay buffer
-            current_z = self.critic(replay_data.observations, replay_data.actions)
-            # Compute critic loss
-            critic_loss = self.quantile_huber_loss(current_z, q_backup)
+                # td error + entropy term
+                target_quantiles = next_quantiles - ent_coef * next_log_prob.reshape(-1, 1)
+                target_quantiles = replay_data.rewards + (1 - replay_data.dones) * self.gamma * target_quantiles
+                # Make it briadcastable with current_quantiles
```
araffin, line 235: typo + please add a comment on what on the shape of target quantiles before and after the transform

sb3_contrib/tqc/tqc.py
```diff
@@ -237,24 +218,27 @@ def train(self, gradient_steps: int, batch_size: int = 64) -> None:
...
+                target_quantiles.unsqueeze_(1)
```
araffin, line 236:
```suggestion
                target_quantiles.unsqueeze_(dim=1)
```

tests/test_cnn.py
```diff
@@ -7,18 +7,18 @@
...
+@pytest.mark.parametrize("model_class", [TQC, QRDQN])
 def test_cnn(tmp_path, model_class):
     SAVE_NAME = "cnn_model.zip"
     # Fake grayscale with frameskip
     # Atari after preprocessing: 84x84x1, here we are using lower resolution
     # to check that the network handle it automatically
     env = FakeImageEnv(screen_height=40, screen_width=40, n_channels=1, discrete=model_class not in {TQC})
     kwargs = {}
-    if model_class in {TQC}:
+    if model_class in {TQC, QRDQN}:
```
araffin, line 21: we should probably also reduce the number of quantiles of QR-QDN during the tests, so they run faster, no?
(QR-DQN uses 200 quantiles by default vs 25 for TQC)

tests/test_cnn.py
```diff
@@ -52,25 +59,27 @@ def params_should_differ(params, other_params):
 @pytest.mark.parametrize("model_class", [TQC])
```
araffin, line 59: not yet included?

tests/test_run.py
```diff
@@ -38,3 +38,17 @@ def test_sde():
     model.learn(total_timesteps=300)
     model.policy.reset_noise()
     model.policy.actor.get_std()
+
+
+def test_qrdqn():
+    model = QRDQN(
+        "MlpPolicy",
+        "CartPole-v1",
+        policy_kwargs=dict(net_arch=[64, 64]),
```
araffin, line 47: dito for the number of quantiles ;)

araffin:
>I pushed some commits to rename them, what do you think?
(I modified TQC's code somewhat to make it easy to understand, sorry about that...)

Overall looks good ;) (I will check in details later)

>Also, how should I evaluate QR-DQN?
The evaluation protocol of sb3 seems different from the paper.
(The paper reported the best agent's score within 125k steps(500k frames))

True, the good news is that we can compare to [Intel Coach](https://github.com/IntelLabs/coach/tree/master/benchmarks/qr_dqn).

I would first compare on simple env (CartPole-v1, LunarLander-v2) and then go for Pong and then Breakout.
For the simpler env, you can probably take a look at [dopamine](https://github.com/google/dopamine) or Intel Coach.
Choose the one which is easier to setup and to tweak, for instance, I did [that](https://github.com/sfujim/TD3/compare/master...araffin:master) (integrating SB3 EvalCallback) to compare SB3 TD3 performance to original implementation.


>Anyway, I really appreciate your kind support ;)


your welcome =)

sb3_contrib/qrdqn/qrdqn.py
ku2482: We sum over a quantile dimension, which is common in QR-DQN, IQN, and FQF.
I'm not sure why they take the mean in TQC.

araffin: It seems that Dopamine is using the mean loss by default: https://github.com/google/dopamine/blob/master/dopamine/jax/agents/quantile/quantile_agent.py#L91
The same goes for IQN implementation of Pfrl: https://github.com/pfnet/pfrl/blob/master/pfrl/agents/iqn.py#L211

or the Facebook ReAgent implementation: https://github.com/facebookresearch/ReAgent/blob/master/reagent/training/qrdqn_trainer.py#L157
maybe that could be a parameter? (and check how it affects the learning)

ku2482: > It seems that Dopamine is using the mean loss by default: https://github.com/google/dopamine/blob/master/dopamine/jax/agents/quantile/quantile_agent.py#L91
The same goes for IQN implementation of Pfrl: https://github.com/pfnet/pfrl/blob/master/pfrl/agents/iqn.py#L211

Their implementations sum over the quantile dimension, so the same as mine, aren't they?
Multiplying "n_quantiles" means that summing over the quantile dimension.

ku2482: Maybe we should add an argument like `sum_over_quantiles : bool` to quantile_huber_loss?

araffin: >Their implementations sum over the (current) quantile dimension, so the same as mine, aren't they?

I will try to check later when I'm rested ;)

>Maybe we should add an argument like sum_over_quantiles : bool to quantile_huber_loss?


Probably, yes

ku2482:
I quickly tested on BreakoutNoFrameskip-v4 to compare with [coach](https://github.com/IntelLabs/coach/tree/master/benchmarks/qr_dqn). (I killed processes due to the limited resource.)
blue: `max_grad_norm=10.0(default)`, orange: `max_grad_norm: np.inf(paper)`

スクリーンショット 2020-12-11 10 23 07

It seems fine compared to coach. So could you help me test QR-DQN in detail?
Here are the things I want to discuss.

- It's redundant if we clip the norm of the grad with `np.inf`. So maybe it's good if `max_grad_norm` can be disabled and we skip clipping? BTW, why DQN in sb3 uses `max_grad_norm=10.0` by default?

- Is it okay that the evaluation protocol for Atari is different from the paper?
I think (i) and (ii) can be applied to sb3.
  1. stop episodes at 27k steps
  2. use epsilon-greedy for evaluation (smaller epsilon than training, like 0.001)
  3. run for 125k steps and report the best episode score.

araffin:
>it's redundant if we clip the norm of the grad with np.inf. So maybe it's good if max_grad_norm can be disabled and we skip clipping? BTW, why DQN in sb3 uses max_grad_norm=10.0 by default?

yes, we can make it optional, especially if it is not mentioned in the paper.
The value comes from OpenAI Baselines implementation.

>Is it okay that the evaluation protocol for Atari is different from the paper?

yes, it should be at least the same as the one we used for checking DQN implementation:
so training for 1e7 steps (and periodic evaluation).

>stop episodes at 27k steps

why 27k steps?

>use epsilon-greedy for evaluation (smaller epsilon than training, like 0.001)

yes, we need to adapt the `EvalCallback` but I will do one run without first.


>EDIT: I used the code below.

Please create a branch in the rl-zoo next time, as mentioned in the contributing guide ;) (I think I will do it)

> Add the argument sum_over_quantiles(or other name)? Should I make cum_prob the optional argument for IQN?

`sum_over_quantiles` sounds good.
For the `cum_prob`, yes, make it probably an optional argument that you compute if not provided.

sb3_contrib/common/utils.py
```diff
@@ -0,0 +1,40 @@
...
+    Partially taken from https://github.com/bayesgroup/tqc_pytorch
+
+    :param current_quantiles: current estimate of quantiles, must be either
+        (batch_size, n_quantiles) or (batch_size, n_critics, n_quantiles)
+    :param target_quantiles: target of quantiles, must be either (batch_size, n_target_quantiles),
+        (batch_size, 1, n_target_quantiles), or (batch_size, n_critics, n_target_quantiles)
+    :return: the loss
+    """
+    if current_quantiles.ndim != target_quantiles.ndim:
+        raise ValueError(
```
araffin, line 16: minor: it would be also good to check inthe tests that `quantile_huber_loss` actually throw those errors when needed.
If necessary, I will do it ;)

ku2482:
> why 27k steps?

[Double DQN paper](https://arxiv.org/abs/1509.06461) mentioned it as below, and most papers follow this evaluation protocol.

>We start an evaluation episode from each of these starting points and run the emulator for up to 108,000 frames (30 mins at 60Hz including the trajectory before the starting point).

> Please create a branch in the rl-zoo next time, as mentioned in the contributing guide ;) (I think I will do it)

Sorry about that, and thank you ;)

araffin:
The RL-Zoo branch is here ;)  https://github.com/DLR-RM/rl-baselines3-zoo/tree/feat/qr-dqn

docs/guide/algos.rst
```diff
@@ -9,6 +9,7 @@ along with some useful characteristics: support for discrete/continuous actions,
 Name         ``Box``     ``Discrete`` ``MultiDiscrete`` ``MultiBinary`` Multi Processing
 ============ =========== ============ ================= =============== ================
 TQC          ✔️          ❌            ❌                ❌              ❌
+QRDQN        ️❌          ✔            ❌                ❌              ❌
```
araffin, line 12: pleas re-use the same character as above (copy-paste should work) and also update the README ;)

docs/guide/examples.rst
```diff
@@ -16,6 +16,22 @@ Train a Truncated Quantile Critics (TQC) agent on the Pendulum environment.
...
+
+Train a Quantile Regression DQN (QR-DQN) agent on the CartPole environment.
+
+.. code-block:: python
+
+  from sb3_contrib import QRDQN
+
+  env = gym.make("CartPole-v1")
+
+  model = QRDQN("MlpPolicy", "CartPole-v1", verbose=1)
```
araffin, line 30:
```suggestion
  model = QRDQN("MlpPolicy", "CartPole-v1", verbose=1)
```

docs/modules/qrdqn.rst
```diff
@@ -0,0 +1,102 @@
...
+Example
+-------
+
+.. code-block:: python
+
+  import gym
+  import numpy as np
+
+  from sb3_contrib import QRDQN
+  from sb3_contrib.qrdqn import MlpPolicy
```
araffin, line 55: Please use the string version (recommended as it is the same for each algorithm) and show how to modify the number of quantiles (cf TQC example in the doc)

docs/guide/examples.rst
```diff
@@ -16,6 +16,22 @@ Train a Truncated Quantile Critics (TQC) agent on the Pendulum environment.
   model.learn(total_timesteps=10000, log_interval=4)
   model.save("tqc_pendulum")
 
+QR-DQN
```
araffin, line 19: Please also update the changelog (put your username next to each entry) and add your username at the bottom ;)

docs/guide/algos.rst
ku2482: You mean, use "QR-DQN", not "QRDQN"?

araffin (commented):
after a quick look, it seems we are only missing the benchmark on Atari and minor edits in the doc, and then we are good for merging =)

I also quickly tested QR-DQN in the zoo on classic control environment (with hyperparameters from DQN) and it works at least as well, so it's a good sign =)

docs/guide/algos.rst
araffin: yes and also the ✔️ seems to be different

ku2482:
I really appreciate your help ;)

docs/modules/qrdqn.rst
```diff
@@ -0,0 +1,101 @@
...
+
+  import gym
+  import numpy as np
+
+  from sb3_contrib import QRDQN
+
+  env = gym.make("CartPole-v1")
+
+  model = QRDQN("MlpPolicy", env, verbose=1)
+  model.learn(total_timesteps=10000, n_quantiles=50, log_interval=4)
```
araffin, line 59: thanks for editing, but this is not how it works, no? (again take a look at TQC example ;) )

ku2482: Sorry about that...

araffin:
My results on 3 seeds (Breakout and Pong) without smoothing window and with the default hyperparameters (so the exploration fraction is wrong and the epsilon for the evaluation is the same as for training):
Results_Breakout
Results_Pong


Overall, it looks good as it matches Intel Coach results (+ a nice boost in final performance ~400 vs ~300 for DQN on Breakout) but as mentioned by @Miffyli that would nice to have at least one run with the original evaluation setting.

ku2482:
Thank you for your support ;)

> but as mentioned by @Miffyli that would nice to have at least one run with the original evaluation setting.

Yeah, but it needs some changes to evaluation callbacks and takes some time.
Maybe, is it good to make another PR to add the evaluation callback for Atari, and add results of DQN, QR-DQN, and other discrete algorithms?

araffin:
> Maybe, is it good to make another PR to add the evaluation callback for Atari, and add results of DQN, QR-DQN, and other discrete algorithms?

yes, probably a good addition for another PR. But I would need @Miffyli advice on that if it is worth implementing ;) (as I'm mostly focused on Continuous Control and not really Atari games with discrete control...)

araffin:
I hope to have more time end of this week or next week to finish the review ;)

Anyway thank you already for your good work =)

sb3_contrib/common/utils.py
```diff
@@ -0,0 +1,59 @@
+from typing import Optional
+
+import torch as th
+
+
+def quantile_huber_loss(
+    current_quantiles: th.Tensor,
+    target_quantiles: th.Tensor,
+    cum_prob: Optional[th.Tensor] = None,
```
araffin, line 9: those are also called "midpoints" in the paper, no?

sb3_contrib/common/utils.py
```diff
@@ -0,0 +1,59 @@
...
+            # For TQC, current_quantiles have a shape (batch_size, n_critics, n_quantiles), and make cum_prob
+            # broadcastable to (batch_size, n_critics, n_quantiles, n_target_quantiles)
+            cum_prob = cum_prob.view(1, 1, -1, 1)
+
+    pairwise_delta = target_quantiles.unsqueeze(-2) - current_quantiles.unsqueeze(-1)
+    abs_pairwise_delta = th.abs(pairwise_delta)
+    huber_loss = th.where(abs_pairwise_delta > 1, abs_pairwise_delta - 0.5, pairwise_delta ** 2 * 0.5)
+    loss = th.abs(cum_prob - (pairwise_delta.detach() < 0).float()) * huber_loss
+    if sum_over_quantiles:
+        loss = loss.sum(dim=-2).mean()
```
araffin, line 56: I think I know why TQC is using the mean directly, this allows to avoid large loss values (and also to have a loss magnitude that does not depends on the number of critics/quantiles),
which makes it probably easier to tune the learning rate.

sb3_contrib/common/utils.py
ku2482: Yes in QR-DQN paper, but it's less common and not used in other papers (IQN, FQF).

sb3_contrib/common/utils.py
ku2482: Probably yes ;)

araffin (approved):
I could finally do the review, LGTM, thanks =)
(only missing to add the benchmark to the doc)

sb3_contrib/common/utils.py
araffin: maybe just add a comment saying "also called midpoints in QR-DQN paper"

ku2482:
Thank you so much ;)

>only missing to add the benchmark to the doc

Could you do this for me, or should I update the doc?

araffin:
>Could you do this for me, or should I update the doc?

done ;)
I also added a small benchmark (5 seeds, tuned hyperparameters) on classic control tasks.

@Miffyli I let you decide if we need more experiments (even though we match Intel Coach results), or we merge ;)

Miffyli:
@araffin 

Visually comparing results here vs. original paper the Breakout and Pong results seem to match up till the trained 10e6, so I would trust this implementation enough to use it as QR-DQN myself and believe we can merge it :).

Just two things: docs should mention what we compared against (Intel Coach and, very roughly, original paper) and also update zoo-instructions once we merge the branch.

ku2482:
@araffin @Miffyli 
Thank you for your kind support ;)

Squeemos:
Hate to post on very late, but is there a specific reason that we cannot use action_noise in QRDQN? Was looking to use it for exploration rather than a exploration scheduler, and would prefer using action noise over noisy nets for now.

Miffyli:
> Hate to post on very late, but is there a specific reason that we cannot use action_noise in QRDQN? Was looking to use it for exploration rather than a exploration scheduler, and would prefer using action noise over noisy nets for now.

QRDQN supports only Discrete action spaces, and there is no action_noise for Discrete spaces (at least no ones implemented in SB3 as of writing).

hh0rva1h:
> My results on 3 seeds (Breakout and Pong) without smoothing window and with the default hyperparameters (so the exploration fraction is wrong and the epsilon for the evaluation is the same as for training):
> Environments 	QRDQN
> 	logs_qrdqn/
> Breakout 	413 +/- 21
> Pong 	20 +/- 0
> 
> Results_Breakout Results_Pong
> 
> Overall, it looks good as it matches Intel Coach results (+ a nice boost in final performance ~400 vs ~300 for DQN on Breakout) but as mentioned by @Miffyli that would nice to have at least one run with the original evaluation setting.

@araffin I tried to reproduce those results, but it did not work out. Which version of the games were used in your benchmark? https://www.gymlibrary.ml/environments/atari/ is documenting 3 versions (v0, v4, v5) and there a couple of different options for the v5 setting as well as different versions of the environment for v4. I tried "Breakout-v5" with default options as well as "BreakoutNoFrameskip-v4" however my learning curves do not look anything comparable to yours.
Code for reproduction (I know I could use `make_atari_env` directly, however for usage with my own monitoring wrappers not included in this example I need to build the environment manually in order to have proper access to the env without vec wrapper and stacking):

```
import gym
#from rl_utils.monitor import Monitor #my custom monitor
from sb3_contrib import QRDQN

from stable_baselines3.common.atari_wrappers import AtariWrapper
from stable_baselines3.common.env_util import make_atari_env
from stable_baselines3.common.vec_env import VecFrameStack, DummyVecEnv

env = gym.make('BreakoutNoFrameskip-v4')
#gym.make('ALE/Breakout-v5', full_action_space=False, frameskip=1)
env = AtariWrapper(env)
env = DummyVecEnv([lambda: env])
env = VecFrameStack(env, n_stack=4)
#env = Monitor(env, params, "./results") # commented out, this is my custom monitor for recording learning curves
#env = make_atari_env(env, n_envs=1, seed=0)
model = QRDQN("CnnPolicy", env, verbose=1,
                             exploration_fraction=0.025,   # match QRDQN PR
                             seed=100)
model.learn(total_timesteps=10000000, log_interval=4)
```
My results looked like the following:

Screenshot from 2022-07-04 10-11-12
Screenshot from 2022-07-04 10-12-24

Would really appreciate if you could give me hints on reproducing your baseline.

araffin:
Hello @hh0rva1h  ,

> Which version of the games were used in your benchmark? 

we always use the NoFrameskip-v4 (and gym 0.21).

> I tried to reproduce those results, but it did not work out. 

Please use the RL Zoo for that, we have instructions in the documentation: 
https://sb3-contrib.readthedocs.io/en/master/modules/qrdqn.html#how-to-replicate-the-results

Instructions are also on huggingface hub: https://huggingface.co/sb3/qrdqn-BreakoutNoFrameskip-v4

and detailed hyperparameters are here: https://github.com/DLR-RM/rl-baselines3-zoo/blob/master/hyperparams/qrdqn.yml#L1


>  however for usage with my own monitoring wrappers not included in this example

I think your issue may come from the monitoring, please read https://github.com/DLR-RM/stable-baselines3/issues/181.

You should at least use `make_vec_env` and pass your custom wrappers.

mreitschuster:
> we always use the NoFrameskip-v4 (and gym 0.21).

any chance you have the reasoning in mind? I thought v4-noframeskip was to be avoided due to the "memorizing sequences" problem (even with up to 30 noops I suspect it just memorizes 31 sequences)

araffin:
>  any chance you have the reasoning in mind?

You can read more about why in the PR on updating gym: 
- https://github.com/DLR-RM/stable-baselines3/pull/572
- https://github.com/DLR-RM/stable-baselines3/pull/734

TL;DR: the new atari v5 were not yet benchmarked and as we have everything using `NoFrameskip-v4`, it is best to keep it like that so we can compare apples to apples.

> due to the "memorizing sequences" problem 

Do you have a reference for that?

mreitschuster:
[Revisiting the Arcade Learning Environment, Machado et al](https://arxiv.org/abs/1709.06009)
**page 7:** _The Arcade Learning Environment is also fully deter-
ministic – each game starts in the same state and outcomes are fully determined by the
state and the action. As such, it is possible to achieve high scores by learning an open-loop
policy, i.e., by simply memorizing a good action sequence, rather than learning to make
good decisions in a variety of game scenarios (Bellemare, Naddaf, Veness, & Bowling, 2015)._
**page 14** -> _Initial Noops ... The environment remains deterministic beyond the choice of starting state. Brute-like methods still perform well._

And to verify: train a model using NoFramskip-v4 and evaluate on v4 (stochasitcity via frameskip) or v5 (stochasticity via stickies) -> it will transfer poorly.
and then other way round: train of v4 or v5 and evaluate it on  NoFramskip-v4 -> good transfer.

I imagine a more sophisitcated approach would be to take randomly some i from 1-30 and use those Noop only in evaluation but not training. If model performs poorly on those frameskips on which it was not trained -> we know it just memorized.

mreitschuster:
@hh0rva1h 

> Do you have by any chance baselines curves at hand for the v4 or v5 scenarios? Would be really great to have more baselines apart from the NoFrameskip-v4. 

Not with qrdqn (yet). working on it. With PPO i have quite a few, but mostly with a non-standard wrapper configuration. But that would go off-topic from this PR. you can have a look at [my work on tuning & env selection](https://github.com/mreitschuster/RLbreakout/tree/master/learning/4_tuning). Short answer: without an aimbot I get to score of 180 and with aimbot to 220 (on 1e7 training steps).I havent found a PM functionality in github - if you want a discussion on that feel free to open an issue/discussion and add me there.

> Also, do you know by any chance how would one exactly match the atari setting as used in the the DQN and QR-DQN papers? ... would be closer to the paper setup than the baselines wrappers, would you agree?
Sorry, no & don't know. I was just testing models on breakout and wanted to check my results with others. Found this thread showing scores of 400 - but then realizing it is in the deterministic environment the excitement level dropped.

araffin:
Thanks @mreitschuster for the links!

> stochasitcity via frameskip

The RL Zoo models are using frameskip (via the `MaxAndSkip` wrapper) but the frame skipping is not stochastic, is that what you meant? (see https://github.com/DLR-RM/stable-baselines3/blob/a6f5049a99a4c21a6f0bcce458ca3306cef310e0/stable_baselines3/common/atari_wrappers.py#L223)

> train a model using NoFrameskip-v4 and evaluate on v4

you need to be careful when you do that, because as mentioned before, if you use atari wrapper, there is frame skipping.

> stochasticity via stickies

by sticky actions you mean randomly repeating actions?

mreitschuster:
@araffin 

> The RL Zoo models are using frameskip (via the MaxAndSkip wrapper) but the frame skipping is not stochastic, is that what you meant?
It is my understanding that we have two different frameskips available - stochastic frameskip (randomly skip frames) which is inbuilt into *-v4 (but deactivated when using *NoFrameskip-v4). it is also not active in v5 (which injects stochasticity diffferently).

and then there is the deterministic frameskip - found in [sb3 atariwrapper](https://stable-baselines3.readthedocs.io/en/master/_modules/stable_baselines3/common/atari_wrappers.html#MaxAndSkipEnv) as well as in the environment [*Deterministic-v4](https://github.com/openai/gym/issues/1280).

> you need to be careful when you do that, because as mentioned before, if you use atari wrapper, there is frame skipping.

yes you are right, but the wrapper provides determinstic frameskip (for speeding up the game), not the stochastic one (to train more transferable skills) 

>by sticky actions you mean randomly repeating actions?
yes. 

 i think if we want to deepen this (and i would be very happy to) it should be a seperate discussion.

PR #18, opened 2021-01-06
Fix the target calculation of QR-DQN

### LINKED ISSUE #17: question regarding qrdqn
Hi,
while trying to extend qrdqn to support double qrdqn, I couldn't convince myself with the current implementation of the update function. 

specifically, refer to the `train` function in  `QRDQN` class (qrdqn.py), line 165:
```python
                # Follow greedy policy: use the one with the highest value
                next_quantiles, _ = next_quantiles.max(dim=2)
```

it looks like the max operation is done on the actions dimensions and by that per each quantile separately, the value from the action that yielded maximum value. (e.g. for the first quantile it can take from a=1, for the 2nd quantile it can take from a=4 etc.)
so the resulting buffer will include quantile values from different actions.

As I have understood (and checked various other implementations of qrdqn), the intent was to calculate the q value of each of the actions by averaging over the quantiles of each action and only then take the quantiles of the best next action. 
something like this:
```python
            with th.no_grad():
                # Compute the quantiles of next observation
                next_quantiles = self.quantile_net_target(replay_data.next_observations)
                best_next_actions = next_quantiles.mean(dim=1).argmax(dim=1,keepdim=True)
                actions_index = best_next_actions[...,None].expand(batch_size,self.n_quantiles,1)
                next_quantiles = target_quantiles.gather(dim=2,index=actions_index).squeeze(dim=2)
                # 1-step TD target
                target_quantiles = replay_data.rewards + (1 - replay_data.dones) * self.gamma * next_quantiles
```

My question : is the current implementation differs by design ? if so, and assuming the implementation that I'm proposing here corresponds to the publication, what's the justification of the variant currently implemented ?

BTW, **Great work with SBL3 !!** a very useful and easy to work with library ! 
I mainly use it (and thus extend it) for offline RL settings. 

Thanks a lot :)

ku2482:
## Description

- [x] Fix the target calculation of QR-DQN(#17)
- [ ] Update results

## Context

closes #17 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:


araffin (commented):
LGTM, please do not forget to update the changelog ;) (you should credit @guyk1971 too)

We will also wait for some benchmark results

EDIT: we should also update the version

guyk1971:
Thanks folks :)
BTW, I have locally applied the update (as specified in the code snippet I
added to the issue) and on preliminary tests (testing on acrobot, cartpole
and mountaincar) the convergence is indeed faster.

Good luck with the important work you're doing !

araffin:
Btw, I think we should add that link to the doc: https://github.com/amy12xx/ml_notes_and_reports/blob/master/distributional_rl/QRDQN.pdf

ku2482:
@araffin 
Sorry for the delay due to the limited resources...
I ran QR-DQN on Breakout and Pong, and the results are almost the same as before.

I apologize for not being able to run with multiple seeds.

- **Breakout**


- **Pong**

araffin:
Hi,

> I ran QR-DQN on Breakout and Pong, and the results are almost the same as before.

I also ran it again and got equivalent results (I just had one bad seed apparently for breakout, hence the worsen results).
I'm surprised that the results does not change that much... (and makes things slightly worse in some cases)

before
breakout
pong
atari

araffin (approved):

araffin:
@ku2482 unless you have some objections, I will merge this once the CI passes ;)

ku2482:
@araffin 
Thank you for evaluating it.

> I just had one bad seed apparently for breakout, hence the worsen results

How are the other seeds? Are they comparable with `QRDQN-before`?

> unless you have some objections, I will merge this once the CI passes ;)

Thanks ;)

araffin:
>How are the other seeds? Are they comparable with QRDQN-before?

I used only 3 seeds in both cases for atari (and 5 for classic control), that's why I don't think the difference is significant.

PR #25, opened 2021-03-30
[Feature] PPO variant with invalid action masking  + maskable common components

### LINKED ISSUE #16: [Feature Request] Invalid action masking
### :rocket: Feature

Add support for masking out invalid actions when making predictions. This would allow models to converge much faster for environments where many actions may be illegal depending on the state.

This feature could be implemented as a wrapper around a gym environment that adds a method to return the mask. The stable baselines algorithms would check for the wrapper and use the mask if available. The mask is a boolean tensor in the shape of the action space, and it replaces the logits for invalid actions with very large negative values in the underlying probability distribution.

[Here is an MVP](https://github.com/kronion/stable-baselines3/compare/master...kronion:feat/action_masking).

### Motivation

In many environments, there may be portions of the action space that are invalid to select in a given state. Without a way to avoid sampling these actions, training becomes less efficient. Models have to waste time exploring the invalid portions of the space, which may become prohibitively expensive for large action spaces. See [this paper](https://arxiv.org/pdf/2006.14171.pdf) for more details.

### Alternatives

To the best of my knowledge, the only alternative is to accept that invalid actions may be selected, and to try to discourage it by penalizing the choice with a large negative reward. This is just the status quo. Note that action masking would be optional, and the status quo would stay the default.

### Additional context

I ran into this problem in practice when building models for board games. To work around it, I implemented an MVP of this feature in a fork. I'd be happy to make a PR. The branch is [here](https://github.com/kronion/stable-baselines3/tree/feat/action_masking).

Also related: hill-a/stable-baselines#453

Original issue here: https://github.com/DLR-RM/stable-baselines3/issues/269 

### Checklist

- [x] I have checked that there is no similar issue in the repo (required)

kronion:
## Description
- Introduces `MaskablePPO`, a variant of SB3's `PPO` algorithm with support for [invalid action masking](https://arxiv.org/abs/2006.14171).
- Introduces various "maskable" subclasses of SB3 data structures needed to implement `MaskablePPO`. These can be reused to build other maskable algorithm variants in the future. Some highlights:
    - `MaskableAlgorithm`, a subclass of `BaseAlgorithm`.
    - `MaskableOnPolicyAlgorithm`, a subclass of `MaskableAlgorithm` that mixes in `OnPolicyAlgorithm`.
    - `MaskablePolicy`, a subclass of `BasePolicy`.
    - `MaskableActorCriticPolicy`, a subclass of `MaskablePolicy` that largely reimplements SB3's `ActorCriticPolicy`.
    - `MaskableRolloutBuffer`, a subclass of `RolloutBuffer` that stores action masks as well.
    - `MaskableCategorical`, a subclass of PyTorch's `Categorical` that gives masked actions a near 0 probability.
- Introduces `ActionMasker` env and venv wrappers. The wrapper takes a function that receives the env as an input and returns the action mask.

Note that a significant amount of code has been ported from SB3 in order to patch classes and methods that would otherwise be fairly internal.

As the checklist shows, there is still work to do for this PR (tests, documentation, etc). I wanted to share what I have so far in order to get early feedback. I could also use guidance/assistance on demonstrating that the functionality is correct. My original intention was to reproduce the experimental results from [this paper](https://arxiv.org/abs/2006.14171), but recreating the control experiment in SB3 has proven challenging.

#### Minimal usage example:
```
import gym
import numpy as np

from sb3_contrib.common.maskable.policies import MaskableActorCriticPolicy
from sb3_contrib.common.wrappers import ActionMasker
from sb3_contrib.ppo_mask import MaskablePPO


def mask_fn(env: gym.Env) -> np.ndarray:
    # Do whatever you'd like in this function to return the action mask
    # for the current env. In this example, we assume the env has a
    # helpful method we can rely on.
    return env.valid_action_mask()


env = ...  # Initialize env
env = ActionMasker(env, mask_fn)  # Wrap to enable masking

# MaskablePPO behaves the same as SB3's PPO unless the env is wrapped
# with ActionMasker. If the wrapper is detected, the masks are automatically
# retrieved and used when learning. Note that MaskablePPO does not accept
# a new action_mask_fn kwarg, as it did in an earlier draft.
model = MaskablePPO(MaskableActorCriticPolicy, env, verbose=1)
model.learn()

# Note that use of masks is manual and optional outside of learning,
# so masking can be "removed" at testing time
model.predict(observation, action_masks=valid_action_array)
```

## Context
Closes #16.
Related to https://github.com/DLR-RM/stable-baselines3/issues/269.

## Types of changes

## Checklist:

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,2 @@
+[mypy-gym.*]
+ignore_missing_imports = True
--- a/sb3_contrib/common/distributions.py
+++ b/sb3_contrib/common/distributions.py
@@ -1,10 +1,19 @@
 from __future__ import annotations
 
-from typing import Optional, Tuple
+from typing import Any, Dict, Optional, Tuple
 
+import gym
 import numpy as np
 import torch as th
-from stable_baselines3.common.distributions import Distribution
+from gym import spaces
+from stable_baselines3.common.distributions import (
+    BernoulliDistribution,
+    DiagGaussianDistribution,
+    Distribution,
+    MultiCategoricalDistribution,
+    StateDependentNoiseDistribution,
+)
+from stable_baselines3.common.preprocessing import get_action_dim
 from torch import nn
 from torch.distributions import Categorical
 
@@ -61,7 +70,7 @@ class CategoricalDistribution(Distribution):
 
     def __init__(self, action_dim: int):
         super(CategoricalDistribution, self).__init__()
-        self.distribution = MaskableCategorical()
+        self.distribution: MaskableCategorical = None
         self.action_dim = action_dim
 
     def proba_distribution_net(self, latent_dim: int) -> nn.Module:
@@ -110,3 +119,36 @@ class CategoricalDistribution(Distribution):
         actions = self.actions_from_params(action_logits)
         log_prob = self.log_prob(actions)
         return actions, log_prob
+
+
+def make_proba_distribution(
+    action_space: gym.spaces.Space, use_sde: bool = False, dist_kwargs: Optional[Dict[str, Any]] = None
+) -> Distribution:
+    """
+    Return an instance of Distribution for the correct type of action space
+
+    :param action_space: the input action space
+    :param use_sde: Force the use of StateDependentNoiseDistribution
+        instead of DiagGaussianDistribution
+    :param dist_kwargs: Keyword arguments to pass to the probability distribution
+    :return: the appropriate Distribution object
+    """
+    if dist_kwargs is None:
+        dist_kwargs = {}
+
+    if isinstance(action_space, spaces.Box):
+        assert len(action_space.shape) == 1, "Error: the action space must be a vector"
+        cls = StateDependentNoiseDistribution if use_sde else DiagGaussianDistribution
+        return cls(get_action_dim(action_space), **dist_kwargs)
+    elif isinstance(action_space, spaces.Discrete):
+        return CategoricalDistribution(action_space.n)
+    elif isinstance(action_space, spaces.MultiDiscrete):
+        return MultiCategoricalDistribution(action_space.nvec)
+    elif isinstance(action_space, spaces.MultiBinary):
+        return BernoulliDistribution(action_space.n)
+    else:
+        raise NotImplementedError(
+            "Error: probability distribution, not implemented for action space"
+            f"of type {type(action_space)}."
+            " Must be of type Gym Spaces: Box, Discrete, MultiDiscrete or MultiBinary."
+        )
--- a/sb3_contrib/common/distributions.py
+++ b/sb3_contrib/common/distributions.py
@@ -1,0 +1,112 @@
+from __future__ import annotations
+
+from typing import Optional, Tuple
+
+import numpy as np
+import torch as th
+from stable_baselines3.common.distributions import Distribution
+from torch import nn
+from torch.distributions import Categorical
+
+
+class MaskableCategorical(Categorical):
+    # Adapted from https://github.com/vwxyzjn/invalid-action-masking
+
+    def __init__(
+        self,
+        probs=None,
+        logits=None,
+        validate_args=None,
+        masks: Optional[np.ndarray] = None,
+    ):
+        self.masks: Optional[th.Tensor] = None
+        super().__init__(probs, logits, validate_args)
+        self.apply_masking(masks)
+
+    def apply_masking(self, masks: Optional[np.ndarray]) -> None:
+        if masks is not None:
+            device = self.logits.device
+            self.masks = th.as_tensor(masks, dtype=th.bool, device=device).reshape(
+                self.logits.shape
+            )
+            HUGE_NEG = th.tensor(-1e8, dtype=self.logits.dtype, device=device)
+
+            logits = th.where(self.masks, self.logits, HUGE_NEG)
+        else:
+            self.masks = None
+            logits = self.logits
+
+        # Reinitialize updated logits
+        super().__init__(logits=logits)
+
+    def entropy(self):
+        if self.masks is None:
+            return super().entropy()
+
+        # Highly negative logits don't result in 0 probs, so we must replace
+        # with 0s to ensure 0 contribution to the distribution's entropy, since
+        # masked actions possess no uncertainty.
+        device = self.logits.device
+        p_log_p = self.logits * self.probs
+        p_log_p = th.where(self.masks, p_log_p, th.tensor(0.0, device=device))
+        return -p_log_p.sum(-1)
+
+
+class CategoricalDistribution(Distribution):
+    """
+    Categorical distribution for discrete actions.
+
+    :param action_dim: Number of discrete actions
+    """
+
+    def __init__(self, action_dim: int):
+        super(CategoricalDistribution, self).__init__()
+        self.distribution = MaskableCategorical()
+        self.action_dim = action_dim
+
+    def proba_distribution_net(self, latent_dim: int) -> nn.Module:
+        """
+        Create the layer that represents the distribution:
+        it will be the logits of the Categorical distribution.
+        You can then get probabilities using a softmax.
+
+        :param latent_dim: Dimension of the last layer
+            of the policy network (before the action layer)
+        :return:
+        """
+        action_logits = nn.Linear(latent_dim, self.action_dim)
+        return action_logits
+
+    def proba_distribution(self, action_logits: th.Tensor) -> CategoricalDistribution:
+        self.distribution = MaskableCategorical(logits=action_logits)
+        return self
+
+    def log_prob(self, actions: th.Tensor) -> th.Tensor:
+        assert self.distribution is not None, "Must set distribution parameters"
+        return self.distribution.log_prob(actions)
+
+    def entropy(self) -> th.Tensor:
+        assert self.distribution is not None, "Must set distribution parameters"
+        return self.distribution.entropy()
+
+    def sample(self) -> th.Tensor:
+        assert self.distribution is not None, "Must set distribution parameters"
+        return self.distribution.sample()
+
+    def mode(self) -> th.Tensor:
+        assert self.distribution is not None, "Must set distribution parameters"
+        return th.argmax(self.distribution.probs, dim=1)
+
+    def actions_from_params(
+        self, action_logits: th.Tensor, deterministic: bool = False
+    ) -> th.Tensor:
+        # Update the proba distribution
+        self.proba_distribution(action_logits)
+        return self.get_actions(deterministic=deterministic)
+
+    def log_prob_from_params(
+        self, action_logits: th.Tensor
+    ) -> Tuple[th.Tensor, th.Tensor]:
+        actions = self.actions_from_params(action_logits)
+        log_prob = self.log_prob(actions)
+        return actions, log_prob
--- a/sb3_contrib/common/maskable/algorithms/base.py
+++ b/sb3_contrib/common/maskable/algorithms/base.py
@@ -1,0 +1,80 @@
+from typing import Optional, Tuple
+
+import numpy as np
+from stable_baselines3.common.base_class import BaseAlgorithm
+from stable_baselines3.common.callbacks import (
+    BaseCallback,
+    CallbackList,
+    ConvertCallback,
+)
+from stable_baselines3.common.type_aliases import MaybeCallback
+from stable_baselines3.common.vec_env import VecEnv
+
+from sb3_contrib.common.maskable.callbacks import MaskableEvalCallback
+from sb3_contrib.common.maskable.policies import MaskablePolicy
+
+
+class MaskableAlgorithm(BaseAlgorithm):
+    def _init_callback(
+        self,
+        callback: MaybeCallback,
+        eval_env: Optional[VecEnv] = None,
+        eval_freq: int = 10000,
+        n_eval_episodes: int = 5,
+        log_path: Optional[str] = None,
+    ) -> BaseCallback:
+        """
+        :param callback: Callback(s) called at every step with state of the algorithm.
+        :param eval_freq: How many steps between evaluations; if None, do not evaluate.
+        :param n_eval_episodes: How many episodes to play per evaluation
+        :param n_eval_episodes: Number of episodes to rollout during evaluation.
+        :param log_path: Path to a folder where the evaluations will be saved
+        :return: A hybrid callback calling `callback` and performing evaluation.
+        """
+        # Convert a list of callbacks into a callback
+        if isinstance(callback, list):
+            callback = CallbackList(callback)
+
+        # Convert functional callback to object
+        if not isinstance(callback, BaseCallback):
+            callback = ConvertCallback(callback)
+
+        # Create eval callback in charge of the evaluation
+        if eval_env is not None:
+            eval_callback = MaskableEvalCallback(
+                eval_env,
+                best_model_save_path=log_path,
+                log_path=log_path,
+                eval_freq=eval_freq,
+                n_eval_episodes=n_eval_episodes,
+            )
+            callback = CallbackList([callback, eval_callback])
+
+        callback.init_callback(self)
+        return callback
+
+    def predict(
+        self,
+        observation: np.ndarray,
+        state: Optional[np.ndarray] = None,
+        mask: Optional[np.ndarray] = None,
+        deterministic: bool = False,
+        action_masks: np.ndarray = None,
+    ) -> Tuple[np.ndarray, Optional[np.ndarray]]:
+        """
+        Get the model's action(s) from an observation
+        :param observation: the input observation
+        :param state: The last states (can be None, used in recurrent policies)
+        :param mask: The last masks (can be None, used in recurrent policies)
+        :param deterministic: Whether or not to return deterministic actions.
+        :param action_masks: Action masks to apply to the action distribution.
+        :return: the model's action and the next state
+            (used in recurrent policies)
+        """
+
+        if isinstance(self.policy, MaskablePolicy):
+            return self.policy.predict(
+                observation, state, mask, deterministic, action_masks=action_masks
+            )
+        else:
+            return self.policy.predict(observation, state, mask, deterministic)
--- a/sb3_contrib/common/maskable/algorithms/base.py
+++ b/sb3_contrib/common/maskable/algorithms/base.py
@@ -8,13 +8,25 @@ from stable_baselines3.common.callbacks import (
     ConvertCallback,
 )
 from stable_baselines3.common.type_aliases import MaybeCallback
-from stable_baselines3.common.vec_env import VecEnv
+from stable_baselines3.common.vec_env import VecEnv, is_vecenv_wrapped
 
 from sb3_contrib.common.maskable.callbacks import MaskableEvalCallback
 from sb3_contrib.common.maskable.policies import MaskablePolicy
+from sb3_contrib.common.vec_env.wrappers import VecActionMasker
+from sb3_contrib.common.wrappers import ActionMasker
 
 
 class MaskableAlgorithm(BaseAlgorithm):
+    @staticmethod
+    def _wrap_env(*args, **kwargs) -> VecEnv:
+        env = super(MaskableAlgorithm, MaskableAlgorithm)._wrap_env(*args, **kwargs)
+
+        # If env is wrapped with ActionMasker, wrap the VecEnv as well for convenience
+        if not is_vecenv_wrapped(env, VecActionMasker) and all(env.env_is_wrapped(ActionMasker)):
+            env = VecActionMasker(env)
+
+        return env
+
     def _init_callback(
         self,
         callback: MaybeCallback,
--- a/sb3_contrib/common/maskable/algorithms/on_policy.py
+++ b/sb3_contrib/common/maskable/algorithms/on_policy.py
@@ -1,0 +1,146 @@
+import gym
+import numpy as np
+import torch as th
+from stable_baselines3.common import logger
+from stable_baselines3.common.buffers import RolloutBuffer
+from stable_baselines3.common.callbacks import BaseCallback
+from stable_baselines3.common.on_policy_algorithm import OnPolicyAlgorithm
+from stable_baselines3.common.vec_env import VecEnv, is_vecenv_wrapped
+
+from sb3_contrib.common.maskable.algorithms.base import MaskableAlgorithm
+from sb3_contrib.common.maskable.buffers import MaskableRolloutBuffer
+from sb3_contrib.common.maskable.policies import MaskablePolicy
+from sb3_contrib.common.vec_env.wrappers import VecActionMasker
+
+
+class MaskableOnPolicyAlgorithm(MaskableAlgorithm, OnPolicyAlgorithm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.rollout_buffer: MaskableRolloutBuffer = None
+
+    def _setup_model(self) -> None:
+        super()._setup_model()
+        self.rollout_buffer = MaskableRolloutBuffer(
+            self.n_steps,
+            self.observation_space,
+            self.action_space,
+            self.device,
+            gamma=self.gamma,
+            gae_lambda=self.gae_lambda,
+            n_envs=self.n_envs,
+        )
+
+        if not isinstance(self.policy, MaskablePolicy):
+            logger.warn("Algorithm's policy does not support invalid action masking")
+
+    def collect_rollouts(
+        self,
+        env: VecEnv,
+        callback: BaseCallback,
+        rollout_buffer: RolloutBuffer,
+        n_rollout_steps: int,
+    ) -> bool:
+        """
+        Collect experiences using the current policy and fill a ``RolloutBuffer``.
+        The term rollout here refers to the model-free notion and should not
+        be used with the concept of rollout used in model-based RL or planning.
+
+        This method is largely identical to the implementation found in the parent class.
+
+        :param env: The training environment
+        :param callback: Callback that will be called at each step
+            (and at the beginning and end of the rollout)
+        :param rollout_buffer: Buffer to fill with rollouts
+        :param n_steps: Number of experiences to collect per environment
+        :return: True if function returned with at least `n_rollout_steps`
+            collected, False if callback terminated rollout prematurely.
+        """
+
+        assert isinstance(
+            rollout_buffer, MaskableRolloutBuffer
+        ), "RolloutBuffer doesn't support action masking"
+        assert self._last_obs is not None, "No previous observation was provided"
+        n_steps = 0
+        action_masks = None
+        rollout_buffer.reset()
+        # Sample new weights for the state dependent exploration
+        if self.use_sde:
+            self.policy.reset_noise(env.num_envs)
+
+        callback.on_rollout_start()
+
+        while n_steps < n_rollout_steps:
+            if (
+                self.use_sde
+                and self.sde_sample_freq > 0
+                and n_steps % self.sde_sample_freq == 0
+            ):
+                # Sample a new noise matrix
+                self.policy.reset_noise(env.num_envs)
+
+            with th.no_grad():
+                # Convert to pytorch tensor
+                obs_tensor = th.as_tensor(self._last_obs).to(self.device)
+
+                # This is the only change related to invalid action masking
+                if is_vecenv_wrapped(env, VecActionMasker):
+                    action_masks = env.valid_actions()
+
+                if isinstance(self.policy, MaskablePolicy):
+                    actions, values, log_probs = self.policy.forward(
+                        obs_tensor, action_masks=action_masks
+                    )
+                else:
+                    actions, values, log_probs = self.policy.forward(obs_tensor)
+
+            actions = actions.cpu().numpy()
+
+            # Rescale and perform action
+            clipped_actions = actions
+            # Clip the actions to avoid out of bound error
+            if isinstance(self.action_space, gym.spaces.Box):
+                clipped_actions = np.clip(
+                    actions, self.action_space.low, self.action_space.high
+                )
+
+            new_obs, rewards, dones, infos = env.step(clipped_actions)
+
+            self.num_timesteps += env.num_envs
+
+            # Give access to local variables
+            callback.update_locals(locals())
+            if callback.on_step() is False:
+                return False
+
+            self._update_info_buffer(infos)
+            n_steps += 1
+
+            if isinstance(self.action_space, gym.spaces.Discrete):
+                # Reshape in case of discrete action
+                actions = actions.reshape(-1, 1)
+
+            rollout_buffer.add(
+                self._last_obs,
+                actions,
+                rewards,
+                self._last_dones,
+                values,
+                log_probs,
+                action_masks,
+            )
+            self._last_obs = new_obs
+            self._last_dones = dones
+
+        with th.no_grad():
+            # Compute value for the last timestep
+            obs_tensor = th.as_tensor(new_obs).to(self.device)
+
+            # Masking is not needed here, the choice of action doesn't matter.
+            # We only want the value of the current observation.
+            _, values, _ = self.policy.forward(obs_tensor)
+
+        rollout_buffer.compute_returns_and_advantage(last_values=values, dones=dones)
+
+        callback.on_rollout_end()
+
+        return True
--- a/sb3_contrib/common/maskable/algorithms/on_policy.py
+++ b/sb3_contrib/common/maskable/algorithms/on_policy.py
@@ -14,10 +14,6 @@ from sb3_contrib.common.vec_env.wrappers import VecActionMasker
 
 
 class MaskableOnPolicyAlgorithm(MaskableAlgorithm, OnPolicyAlgorithm):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.rollout_buffer: MaskableRolloutBuffer = None
-
     def _setup_model(self) -> None:
         super()._setup_model()
         self.rollout_buffer = MaskableRolloutBuffer(
@@ -126,7 +122,7 @@ class MaskableOnPolicyAlgorithm(MaskableAlgorithm, OnPolicyAlgorithm):
                 self._last_dones,
                 values,
                 log_probs,
-                action_masks,
+                action_masks=action_masks,
             )
             self._last_obs = new_obs
             self._last_dones = dones
--- a/sb3_contrib/common/maskable/buffers.py
+++ b/sb3_contrib/common/maskable/buffers.py
@@ -1,4 +1,4 @@
-from typing import Generator, Optional
+from typing import Generator, NamedTuple, Optional
 
 import numpy as np
 import torch as th
@@ -8,8 +8,17 @@ from stable_baselines3.common.type_aliases import RolloutBufferSamples
 from stable_baselines3.common.vec_env import VecNormalize
 
 
-class MaskableRolloutBufferSamples(RolloutBufferSamples):
-    action_masks: th.Tensor
+# NamedTuple doesn't support straightforward subclassing, but it can be
+# achieved with some hacks. TODO use dataclasses instead?
+# Based on https://stackoverflow.com/a/62160225
+# Depends on python 3.6+ to avoid field order changing.
+_sample_types = list(RolloutBufferSamples.__annotations__.items()) + [("action_masks", th.Tensor)]
+_maskable_samples = NamedTuple("_maskable_samples", _sample_types)  # mypy dislikes non-literal list
+
+
+# Mix the original NamedTuple in over again, in order to get desired issubclass() behavior
+class MaskableRolloutBufferSamples(_maskable_samples, RolloutBufferSamples):
+    pass
 
 
 class MaskableRolloutBuffer(RolloutBuffer):
@@ -52,11 +61,11 @@ class MaskableRolloutBuffer(RolloutBuffer):
             for tensor in [
                 "observations",
                 "actions",
-                "action_masks",
                 "values",
                 "log_probs",
                 "advantages",
                 "returns",
+                "action_masks",
             ]:
                 self.__dict__[tensor] = self.swap_and_flatten(self.__dict__[tensor])
             self.generator_ready = True
@@ -76,10 +85,10 @@ class MaskableRolloutBuffer(RolloutBuffer):
         data = (
             self.observations[batch_inds],
             self.actions[batch_inds],
-            self.action_masks[batch_inds].flatten(),
             self.values[batch_inds].flatten(),
             self.log_probs[batch_inds].flatten(),
             self.advantages[batch_inds].flatten(),
             self.returns[batch_inds].flatten(),
+            self.action_masks[batch_inds].flatten(),
         )
         return MaskableRolloutBufferSamples(*tuple(map(self.to_torch, data)))
--- a/sb3_contrib/common/maskable/buffers.py
+++ b/sb3_contrib/common/maskable/buffers.py
@@ -1,0 +1,85 @@
+from typing import Generator, Optional
+
+import numpy as np
+import torch as th
+from gym import spaces
+from stable_baselines3.common.buffers import RolloutBuffer
+from stable_baselines3.common.type_aliases import RolloutBufferSamples
+from stable_baselines3.common.vec_env import VecNormalize
+
+
+class MaskableRolloutBufferSamples(RolloutBufferSamples):
+    action_masks: th.Tensor
+
+
+class MaskableRolloutBuffer(RolloutBuffer):
+    def __init__(self, *args, **kwargs):
+        self.action_masks = None
+        super().__init__(*args, **kwargs)
+
+    def reset(self) -> None:
+        # TODO: need to handle different action_space attrs based on shape
+        action_masks = np.array([], dtype=np.float32)
+        if isinstance(self.action_space, spaces.Discrete):
+            action_masks = np.ones(
+                (self.buffer_size, self.n_envs, self.action_space.n), dtype=np.float32
+            )
+        elif isinstance(self.action_space, spaces.MultiDiscrete):
+            action_masks = np.ones(
+                (self.buffer_size, self.n_envs, sum(self.action_space.nvec)),
+                dtype=np.float32,
+            )
+        self.action_masks = action_masks
+
+        super().reset()
+
+    def add(self, *args, action_masks: np.ndarray = None, **kwargs) -> None:
+        """
+        :param action_masks: Masks applied to constrain the choice of possible actions.
+        """
+        if action_masks is not None:
+            self.action_masks[self.pos] = np.array(action_masks).copy()
+
+        super().add(*args, **kwargs)
+
+    def get(
+        self, batch_size: Optional[int] = None
+    ) -> Generator[MaskableRolloutBufferSamples, None, None]:
+        assert self.full, ""
+        indices = np.random.permutation(self.buffer_size * self.n_envs)
+        # Prepare the data
+        if not self.generator_ready:
+            for tensor in [
+                "observations",
+                "actions",
+                "action_masks",
+                "values",
+                "log_probs",
+                "advantages",
+                "returns",
+            ]:
+                self.__dict__[tensor] = self.swap_and_flatten(self.__dict__[tensor])
+            self.generator_ready = True
+
+        # Return everything, don't create minibatches
+        if batch_size is None:
+            batch_size = self.buffer_size * self.n_envs
+
+        start_idx = 0
+        while start_idx < self.buffer_size * self.n_envs:
+            yield self._get_samples(indices[start_idx : start_idx + batch_size])
+            start_idx += batch_size
+
+    def _get_samples(
+        self, batch_inds: np.ndarray, env: Optional[VecNormalize] = None
+    ) -> MaskableRolloutBufferSamples:
+        data = (
+            self.observations[batch_inds],
+            self.actions[batch_inds],
+            self.action_masks[batch_inds].flatten(),
+            self.values[batch_inds].flatten(),
+            self.log_probs[batch_inds].flatten(),
+            self.advantages[batch_inds].flatten(),
+            self.returns[batch_inds].flatten(),
+        )
+        return MaskableRolloutBufferSamples(*tuple(map(self.to_torch, data)))
--- a/sb3_contrib/common/vec_env/wrappers/action_masker.py
+++ b/sb3_contrib/common/vec_env/wrappers/action_masker.py
@@ -1,0 +1,35 @@
+from typing import Callable, List
+
+import gym
+import numpy as np
+from stable_baselines3.common.vec_env import VecEnv, VecEnvWrapper
+from stable_baselines3.common.vec_env.base_vec_env import VecEnvObs, VecEnvStepReturn
+
+from sb3_contrib.common.wrappers import ActionMasker
+
+
+class VecActionMasker(VecEnvWrapper):
+    """
+    A wrapper for a VecEnv of envs that have themselves been wrapped with the ActionMasker wrapper.
+
+    :param venv: the VecEnv to wrap. The underlying envs must be wrapped with the ActionMasker wrapper.
+    """
+
+    def __init__(self, venv: VecEnv):
+        super().__init__(venv)
+
+        if not all(self.env_is_wrapped(ActionMasker)):
+            raise ValueError("VecEnv's envs must be wrapped with the ActionMasker wrapper")
+
+    def valid_actions(self) -> List[np.ndarray]:
+        return self.venv.env_method("valid_actions")
+
+    def reset(self) -> VecEnvObs:
+        return self.venv.reset()
+
+    def step_wait(self) -> VecEnvStepReturn:
+        return self.venv.step_wait()
+
+    def action_mask_fn(self) -> Callable[[gym.Env], np.ndarray]:
+        # Assumes that all environments use the same masking function.
+        return self.get_attr("action_mask_fn")[0]
--- a/sb3_contrib/common/wrappers/action_masker.py
+++ b/sb3_contrib/common/wrappers/action_masker.py
@@ -1,0 +1,33 @@
+from typing import Callable, Union
+
+import gym
+import numpy as np
+
+
+class ActionMasker(gym.Wrapper):
+    """
+    Vectorized environment base class
+
+    :param env: the Gym environment to wrap
+    :param action_mask_fn: A function that takes a Gym environment and returns an action mask,
+        or the name of such a method provided by the environment.
+    """
+
+    def __init__(
+        self, env: gym.Env, action_mask_fn: Union[str, Callable[[gym.Env], np.ndarray]]
+    ):
+        super().__init__(env)
+
+        if isinstance(action_mask_fn, str):
+            found_method = getattr(self.env, action_mask_fn)
+            if not callable(found_method):
+                raise ValueError(
+                    f"Environment attribute {action_mask_fn} is not a method"
+                )
+
+            self.action_action_mask_fn = found_method
+        else:
+            self.action_mask_fn = action_mask_fn
+
+    def valid_actions(self) -> np.ndarray:
+        return self.action_mask_fn(self.env)
--- a/sb3_contrib/ppo_mask/ppo_mask.py
+++ b/sb3_contrib/ppo_mask/ppo_mask.py
@@ -1,0 +1,149 @@
+import numpy as np
+import torch as th
+from gym import spaces
+from stable_baselines3.common import logger
+from stable_baselines3.common.utils import explained_variance
+from stable_baselines3.ppo import PPO
+from torch.nn import functional as F
+
+from sb3_contrib.common.maskable.algorithms.on_policy import MaskableOnPolicyAlgorithm
+from sb3_contrib.common.maskable.policies import MaskablePolicy
+
+
+class MaskablePPO(MaskableOnPolicyAlgorithm, PPO):
+    def train(self) -> None:
+        """
+        Update policy using the currently gathered rollout buffer.
+        """
+        # Update optimizer learning rate
+        self._update_learning_rate(self.policy.optimizer)
+        # Compute current clip range
+        clip_range = self.clip_range(self._current_progress_remaining)
+        # Optional: clip range for the value function
+        if self.clip_range_vf is not None:
+            clip_range_vf = self.clip_range_vf(self._current_progress_remaining)
+
+        entropy_losses, all_kl_divs = [], []
+        pg_losses, value_losses = [], []
+        clip_fractions = []
+
+        # train for n_epochs epochs
+        for epoch in range(self.n_epochs):
+            approx_kl_divs = []
+            # Do a complete pass on the rollout buffer
+            for rollout_data in self.rollout_buffer.get(self.batch_size):
+                actions = rollout_data.actions
+                if isinstance(self.action_space, spaces.Discrete):
+                    # Convert discrete action from float to long
+                    actions = rollout_data.actions.long().flatten()
+
+                # Re-sample the noise matrix because the log_std has changed
+                # TODO: investigate why there is no issue with the gradient
+                # if that line is commented (as in SAC)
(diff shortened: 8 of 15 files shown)


sb3_contrib/common/distributions.py
```diff
@@ -0,0 +1,146 @@
...
+    Distribution,
+    MultiCategoricalDistribution,
+    StateDependentNoiseDistribution,
+)
+from stable_baselines3.common.preprocessing import get_action_dim
+from torch import nn
+from torch.distributions import Categorical
+
+
+class MaskableCategorical(Categorical):
```
araffin, line 19: can't you merge this one with `CategoricalDistribution` ?

sb3_contrib/common/distributions.py
```diff
@@ -0,0 +1,146 @@
...
+        self.proba_distribution(action_logits)
+        return self.get_actions(deterministic=deterministic)
+
+    def log_prob_from_params(self, action_logits: th.Tensor) -> Tuple[th.Tensor, th.Tensor]:
+        actions = self.actions_from_params(action_logits)
+        log_prob = self.log_prob(actions)
+        return actions, log_prob
+
+
+def make_proba_distribution(
```
araffin, line 116: this should not be needed as `PPO_MASK` will only use `MaskedCategorical` distribution, no?

sb3_contrib/common/maskable/algorithms/base.py
```diff
@@ -0,0 +1,86 @@
...
+from stable_baselines3.common.type_aliases import MaybeCallback
+from stable_baselines3.common.vec_env import VecEnv, is_vecenv_wrapped
+
+from sb3_contrib.common.maskable.callbacks import MaskableEvalCallback
+from sb3_contrib.common.maskable.policies import MaskablePolicy
+from sb3_contrib.common.vec_env.wrappers import VecActionMasker
+from sb3_contrib.common.wrappers import ActionMasker
+
+
+class MaskableAlgorithm(BaseAlgorithm):
```
araffin, line 15: you don't need to define a new base class, you can overwrite `wrap_env` in the PPO algorithm only

sb3_contrib/common/maskable/algorithms/on_policy.py
```diff
@@ -0,0 +1,132 @@
...
+from stable_baselines3.common.on_policy_algorithm import OnPolicyAlgorithm
+from stable_baselines3.common.vec_env import VecEnv, is_vecenv_wrapped
+
+from sb3_contrib.common.maskable.algorithms.base import MaskableAlgorithm
+from sb3_contrib.common.maskable.buffers import MaskableRolloutBuffer
+from sb3_contrib.common.maskable.policies import MaskablePolicy
+from sb3_contrib.common.vec_env.wrappers import VecActionMasker
+
+
+class MaskableOnPolicyAlgorithm(MaskableAlgorithm, OnPolicyAlgorithm):
```
araffin, line 16: same remark as before, this can be done directly in PPO

sb3_contrib/common/maskable/buffers.py
```diff
@@ -0,0 +1,87 @@
+from typing import Generator, NamedTuple, Optional
+
+import numpy as np
+import torch as th
+from gym import spaces
+from stable_baselines3.common.buffers import RolloutBuffer
+from stable_baselines3.common.type_aliases import RolloutBufferSamples
+from stable_baselines3.common.vec_env import VecNormalize
+
+# NamedTuple doesn't support straightforward subclassing, but it can be
```
araffin, line 10: why not re-defining `MaskableRolloutBufferSamples` completely, without deriving from `RolloutBufferSamples` ?

sb3_contrib/common/vec_env/wrappers/action_masker.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+
+    def __init__(self, venv: VecEnv):
+        super().__init__(venv)
+
+        if not all(self.env_is_wrapped(ActionMasker)):
+            raise ValueError("VecEnv's envs must be wrapped with the ActionMasker wrapper")
+
+    def valid_actions(self) -> List[np.ndarray]:
+        return self.venv.env_method("valid_actions")
```
araffin, line 25: is this wrapper really needed?
or is it just to make things look nicer?

because it looks like you could just define two helper functions, `valid_actions()` and `action_mask_fn()` without having to redefine a wrapper

sb3_contrib/common/wrappers/action_masker.py
```diff
@@ -0,0 +1,29 @@
+from typing import Callable, Union
+
+import gym
+import numpy as np
+
+
+class ActionMasker(gym.Wrapper):
```
araffin, line 7: instead of this wrapper, could we just require the user to provide an action mask function with a specific name? (this would simplify things, no?)

sb3_contrib/ppo_mask/ppo_mask.py
```diff
@@ -0,0 +1,128 @@
...
+                    # Convert discrete action from float to long
+                    actions = rollout_data.actions.long().flatten()
+
+                # Re-sample the noise matrix because the log_std has changed
+                # TODO: investigate why there is no issue with the gradient
+                # if that line is commented (as in SAC)
+                if self.use_sde:
+                    self.policy.reset_noise(self.batch_size)
+
+                if isinstance(self.policy, MaskablePolicy):
```
araffin, line 46: you don't need that if as `PPO_MASK` will be used with `MaskablePolicy` only

araffin (commented):
thanks for the PR =)
I've written some quick comments that should simplify the implementation ;)

araffin:
> My original intention was to reproduce the experimental results from this paper, but recreating the control experiment in SB3 has proven challenging.

What is the challenging part? The use of dictionary observation?

Two solutions:
- flatten the dict with a wrapper if possible
- use https://github.com/DLR-RM/stable-baselines3/pull/243 (the PR is ready, mostly missing additional doc and tests)

sb3_contrib/common/maskable/algorithms/on_policy.py
```diff
@@ -0,0 +1,132 @@
...
+                else:
+                    actions, values, log_probs = self.policy.forward(obs_tensor)
+
+            actions = actions.cpu().numpy()
+
+            # Rescale and perform action
+            clipped_actions = actions
+            # Clip the actions to avoid out of bound error
+            if isinstance(self.action_space, gym.spaces.Box):
+                clipped_actions = np.clip(actions, self.action_space.low, self.action_space.high)
```
vwxyzjn, line 90: Saw some code for the continuous action space as well. However, it seems the current implementation only handles the discrete action space. Also I have no idea how to implement invalid action masking with continuous action space haha.

Maybe consider throwing an error if the action space is continuous?

vwxyzjn:
Thanks for pining me @araffin. Overall I think @kronion did a great job in implementing the core idea of invalid action masking: use the mask during sampling *and* training. 

Is there any way to test run this? The minimal example seems to require a bit more code on the testing environment to run.

I am going to be quite busy for the next week, but hopefully will be able to get some time to further contribute to this conversation.

sb3_contrib/common/distributions.py
kronion: Yes, but users could provide environments with any space they'd like. I wrote the algorithm as though masking was an optional opt-in feature on top of the existing `PPO`, instead of raising an exception if the user used the algorithm in a way that wouldn't result in masking.

I guess that's overly permissive, it's probably clearer to just raise the exception. If the user is happy to proceed without masking, they should just use SB3's `PPO`. If this ever became a native SB3 feature, it would be a different story.

That said, I'd like to add other distributions besides `Categorical`, at least `MultiCategorical` and `Bernoulli`. So this function would probably be useful in the future.

sb3_contrib/common/distributions.py
kronion: How so? Do you mean combining into a single class, or just moving this to be an inner class?

sb3_contrib/common/maskable/algorithms/base.py
kronion: Wouldn't it be helpful to have a base class that could be used for other algorithms besides PPO? Even if that's out of scope right now, is that not something worth supporting as a possibility down the line?

sb3_contrib/common/maskable/algorithms/on_policy.py
kronion: Same response as above, is it better to design for future extension?

sb3_contrib/common/vec_env/wrappers/action_masker.py
kronion: Yes, it's mostly syntactic sugar. We could make do with a helper around calling `venv.env_method("valid_actions")`.

The other use for this class is to indicate that masking is desired, but there's no need for that if it's an exception to use the algorithm without masking.

kronion (commented):
I tried to reproduce the experiment results in this repo: https://github.com/vwxyzjn/invalid-action-masking

SB3 wasn't used, so I started out by trying to rewrite the control experiment (ie no masking) to use SB3's `PPO`, so I had a baseline to compare against. This code specifically: https://github.com/vwxyzjn/invalid-action-masking/blob/master/invalid_action_masking/ppo_no_mask_4x4.py

Unfortunately I wasn't successful, it seemed that SB3's `PPO` converged far slower than the results in the paper. When I ran the original code, the model would converge to an average episode score near 40 by 100k timesteps (figure 3a in [the paper](https://arxiv.org/pdf/2006.14171.pdf#cite.sutton2000policy)). I was seeing far lower scores by a 100k cutoff, I think closer to 5.

Most likely I simply failed to rewrite the net correctly. I just ran out of time and haven't had a chance to revisit it. I'll create a repo for that work and share it here shortly.

As an aside, I haven't run into dictionary observations while working on this. Should I have?

kronion:
> Is there any way to test run this? The minimal example seems to require a bit more code on the testing environment to run.

Not yet, unfortunately. It would be great to demonstrate how it works on a small example env, but I don't have lying around. Does anybody have an env they'd suggest?

sb3_contrib/common/vec_env/wrappers/action_masker.py
araffin: >The other use for this class is to indicate that masking is desired, but there's no need for that if it's an exception to use the algorithm without masking.

if the user uses `PPO_MASK`, I hope that he/she knows that masking is required

sb3_contrib/common/wrappers/action_masker.py
araffin: >You mean something along the lines of hasattr(env, "valid_actions")?

yes, the same way `compute_reward()` is required for `gym.GoalEnv`, we would force the user to implement an action mask function. It is a minimal requirement that should simplify a lot the code.

>but I still expect many users may need to use this wrapper on existing environments in order to provide the expected method.

we could document it (as it is only 5 lines of code if you remove the existence check) but you are right, we could just leave it without using it directly.

Makefile
```diff
@@ -24,7 +24,7 @@ check-codestyle:
 	# Sort imports
 	isort --check ${LINT_PATHS}
 	# Reformat using black
-	black --check -l 127 ${LINT_PATHS}
+	black --check ${LINT_PATHS}
```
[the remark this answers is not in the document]
kronion: Just as a demonstration that it wasn't needed if `pyproject.toml` is included.

mypy.ini
```diff
@@ -0,0 +1,2 @@
+[mypy-gym.*]
```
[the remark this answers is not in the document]
kronion: Fair enough!

pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[tool.black]
```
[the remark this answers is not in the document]
kronion: Yes. Typically IDE plugins simply locate and call the `black` executable, which then looks for configuration defined in this file. Specifying the desired length only in the Makefile means that `black` doesn't do the right thing in general. This bit me, I was depending on an automatic vim plugin rather than calling the Makefile directly.

The Makefile works the same as before, though, because `black` finds the config file when it's called.

It's nice that `make format` is there, but it would be even better if `black` did the desired thing no matter how it was called (IDE, command line, in the Makefile).

sb3_contrib/common/maskable/algorithms/base.py
kronion: I was thinking along the lines of other masked algorithms just within this repo, but it's hard to argue with starting simple :grinning:

benallard:
> > Is there any way to test run this? The minimal example seems to require a bit more code on the testing environment to run.
> 
> Not yet, unfortunately. It would be great to demonstrate how it works on a small example env, but I don't have lying around. Does anybody have an env they'd suggest?

I'd be glad to propose my small toy-environment [`1PQuarto`](https://github.com/benallard/quarto-gym) for that purpose. It's an Environment about the game [Quarto](https://en.wikipedia.org/wiki/Quarto_(board_game)). the 1P adds a random player so that the agent always plays from the same side. Each move consist of two parts, and both have reasons to be invalid. I've been rewarding negative rewards for invalid move at the moment, but that doesn't word as I'd like, hence me poking around what other solution I could have for that.

Note that I'm still very new to all that, so that example might be quite bad for various reasons I don't understand at the moment.

sb3_contrib/common/distributions.py
kronion: It was indeed reusable, so resolving for now.

sb3_contrib/common/maskable/buffers.py
```diff
@@ -0,0 +1,85 @@
...
+        self.action_masks = action_masks
+
+        super().reset()
+
+    def add(self, *args, action_masks: np.ndarray = None, **kwargs) -> None:
+        """
+        :param action_masks: Masks applied to constrain the choice of possible actions.
+        """
+        if action_masks is not None:
+            self.action_masks[self.pos] = np.array(action_masks).copy()
```
benallard, line 45:
```suggestion
            self.action_masks[self.pos] = np.array(action_masks).reshape(1, sum(self.action_space.nvec)).copy()
```

Was necessary for me.

kronion:
@benallard Thanks for the env suggestion, and the review. Quarto is a good choice for demonstrating how masking works in a multidiscrete action space. It's also an easy game to understand and visualize.

Per earlier discussion, I've started to work on making an easy-to-run demonstration of how the feature works. I used Quarto. I made a separate repo for it: https://github.com/kronion/sb3-action-masking-demo. I'll share more results soon, but for now I've included instructions there for how to run a few simple experiments between two agents.

sb3_contrib/common/maskable/utils.py
```diff
@@ -0,0 +1,31 @@
...
+        if isinstance(env, VecEnv):
+            return env.get_attr(attr_name)[0]
+        else:
+            return getattr(env, attr_name)
+    except AttributeError:
+        return None
+
+
+def is_masking_supported(env: GymEnv) -> bool:
+    return get_action_mask_fn(env) is not None
```
benallard, line 31: I'd thought looking for the `action_masks` method would be enough (no need to call it, just make sure it's there). Do you really need to check for the `fn`  there, and latter on just use the other method ? There must be a way to go with just one.

At the moment if I only define `action_masks`, it (seems to) work properly, but spit a (wrong?) warning about my Env not supporting masking ...

benallard: Okay, so it seems you need the `fn` to wrap other Envs (like the evaluation one in `MaskableEvalCallback`).

In My env I just defined

``` py
class MyEnv(gym.Env):
    ...
    def action_masks(self):
        ...
    action_mask_fn = action_masks
```

And that did it. So one of them really seems to be too much.

sb3_contrib/common/maskable/buffers.py
```diff
@@ -0,0 +1,85 @@
...
+        self.action_masks = action_masks
+
+        super().reset()
+
+    def add(self, *args, action_masks: np.ndarray = None, **kwargs) -> None:
+        """
+        :param action_masks: Masks applied to constrain the choice of possible actions.
+        """
+        if action_masks is not None:
+            self.action_masks[self.pos] = action_masks.flatten()
```
kronion, line 45: @benallard can you confirm that this change works for you? I wasn't able to use your suggestion because not all gym action spaces have the `nvec` attribute, it's specfic to multidiscrete.

araffin: not sure what was the issue (what is the expected shape?)... maybe you can take a look at https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/buffers.py#L364

and 
https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/buffers.py#L45 defined https://github.com/DLR-RM/stable-baselines3/blob/ddbe0e93f9fe55152f2354afd058b28e6ccc3345/stable_baselines3/common/preprocessing.py#L161

but I assume you are expecting the action mask to be already preprocessed? (one hot vector)

kronion: The issue was that the code originally expected the mask array to be a 1D array, regardless of the shape of the action space:
```
        if action_masks is not None:
            self.action_masks[self.pos] = np.array(action_masks).copy()
```
This was a surprise to benallard. He tried to pass a mask with a separate array per choice in his multidiscrete action space:
```
def mask_fn(env: gym.Env) -> np.ndarray:
    positions = set(a[0] for a in env.legal_actions)
    pieces = set(a[1] for a in env.legal_actions)
    action_mask = [[], []]  # Separate array per choice in multidiscrete
    for i in range(16):
        action_mask[0].append(i in positions)
        action_mask[1].append(i in pieces)
    return action_mask
```

I thought it would be better not to be strict about the shape of the provided mask and simply flatten it internally. We could manually reshape to match our expectation, per benallard's earlier suggestion:
```
self.action_masks[self.pos] = np.array(action_masks).reshape(1, sum(self.action_space.nvec)).copy()
```
The approach for doing this would differ based the kind of action space used. I thought it wasn't necessary because the mask is reshaped again [when it's applied to the distribution](https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/pull/25/files#diff-6a50138bae28ed7a66f6feda0989635797e3ee4e5adaa0d78ebc63b8da5ada3bR180-R184). If the shape is wrong, it would be caught then.

benallard: I can confirm that it works 👍

minhlong94:
@kronion small comment, you make a typo there. It is `MaskedActorCriticPolicy` instead of `MaskableActorCriticPolicy`. Same thing with `MaskablePPO`.

kronion:
Apologies for the naming confusion. I've gone back and forth on whether `Maskable` or `Masked` is better. Currently I think `Maskable` is more accurate because you can use the algorithm with masking disabled (at test time, or in an environment where masks aren't available, etc).

I expect the maintainers will have opinions on naming as well, so nothing is canonical yet.

Miffyli:
I would say `Maskable`, despite being a bit weird word, is more suitable because of the "it can be masked, but not necessary". Either or works for me, and at this point I vote for less unnecessary work :).

By the way, are you expecting further reviews or comments or is something WIP?

kronion:
I don't have anything WIP at this point. I've left a few review comments unresolved pending additional feedback or a :+1: from a reviewer, but I'm hoping they can mostly be marked resolved. Further review would be welcome,

I think the biggest outstanding item is validating functionality/performance. I've been working on fleshing out [this demo](https://github.com/kronion/sb3-action-masking-demo) to make it easier for try out the feature and verify that it behaves as desired.

Miffyli:
>I think the biggest outstanding item is validating functionality/performance. I've been working on fleshing out this demo to make it easier for try out the feature and verify that it behaves as desired.

Ok sounds good 👍 . If you can finish up this part and provide some results on what is expected and what this obtains, then we are very close to merging. Remember to include instructions on replicating the results in the docs (see other algos) :)

araffin:
>I don't have anything WIP at this point. I've left a few review comments unresolved pending additional feedback or a +1 from a reviewer, but I'm hoping they can mostly be marked resolved. Further review would be welcome,

Thanks for letting us know =)

I don't have much time this week for an additional review (maybe @Miffyli has), otherwise I will try to do it next week ;)

Miffyli:
> I don't have much time this week for an additional review (maybe @Miffyli has), otherwise I will try to do it next week ;)

Gah I forgot to reply that I too won't have too much time to share right now (deadlines, despite being vanquished at a steady pace, never seem to go extinct).

qorrect:
Looking forward to this getting merged, thanks all.

araffin:
@kronion before I do a complete review, it seems that documentation and tests are missing... (cf checklist and contributing guide)
Could you add that?  (or can someone help you?)

zachoooo:
Thanks for your work on this. I've managed to get this to run on my system but it doesn't seem to work when `n_envs` is set to a value other than 1.

kronion:
@araffin will do, but I'll likely need a few weeks to complete that work.

I've made some progress on running tests of performance on the game of quarto. I've checked in the tensorboard results of two 4M timestep training sessions, one with masking and one without. [They're in this repo](https://github.com/kronion/sb3-action-masking-demo). If you clone it, you can view the results with this command:
```
tensorboard --logdir_spec vanilla:zoo/long/refactored/latest/PPO_1,masked:zoo/long_mask/refactored/latest/PPO_1
```

I would really appreciate it if somebody could glance through the tensorboard results and double-check whether any of these seem amiss. Many of the metric differences seem reasonable/plausible to me, but I'm concerned about `explained_variance`. I guess the value net is making predictions as if there's no masking, and this leads to poor predictions and high variance? I'm not quite sure.

@zachoooo thanks for the report. Will look into it once a few of these other items have been addressed.

zachoooo:
> @araffin will do, but I'll likely need a few weeks to complete that work.
> 
> I've made some progress on running tests of performance on the game of quarto. I've checked in the tensorboard results of two 4M timestep training sessions, one with masking and one without. [They're in this repo](https://github.com/kronion/sb3-action-masking-demo). If you clone it, you can view the results with this command:
> 
> ```
> tensorboard --logdir_spec vanilla:zoo/long/refactored/latest/PPO_1,masked:zoo/long_mask/refactored/latest/PPO_1
> ```
> 
> I would really appreciate it if somebody could glance through the tensorboard results and double-check whether any of these seem amiss. Many of the metric differences seem reasonable/plausible to me, but I'm concerned about `explained_variance`. I guess the value net is making predictions as if there's no masking, and this leads to poor predictions and high variance? I'm not quite sure.
> 
> @zachoooo thanks for the report. Will look into it once a few of these other items have been addressed.

I'm actually not really sure if this is just an issue with my environment or not. If I wrap my environment in a DummyVecWrapper then my mask function isn't able to access methods and variables in the environment. I might be able to fix this by using a wrapper API call to call the function instead of trying to call it directly. I'm pretty busy lately, but I'll try to look into this when I have some time.

sb3_contrib/ppo_mask/ppo_mask.py
```diff
@@ -0,0 +1,438 @@
...
+            _init_setup_model=False,
+            supported_action_spaces=(
+                spaces.Discrete,
+                spaces.MultiDiscrete,
+                spaces.MultiBinary,
+            ),
+        )
+
+        if use_masking and not is_masking_supported(self.env):
+            logger.warn("Environment does not support action masking. Consider using ActionMasker wrapper")
```
araffin, line 116: I would throw an error, no?

sb3_contrib/ppo_mask/ppo_mask.py
```diff
@@ -0,0 +1,438 @@
...
+
+            if isinstance(self.action_space, spaces.Discrete):
+                # Reshape in case of discrete action
+                actions = actions.reshape(-1, 1)
+
+            rollout_buffer.add(
+                self._last_obs,
+                actions,
+                rewards,
+                self._last_dones,
```
araffin, line 267: in the most recent SB3 version, this was renamed to `self._last_episode_starts` to better reflect its role, see https://github.com/DLR-RM/stable-baselines3/pull/375

kronion: Is there a good way for me to account for that now, or should I wait for v1.1 to be released?

sb3_contrib/ppo_mask/ppo_mask.py
araffin: why would someone do that?
There is normal PPO for such env, no?

pyproject.toml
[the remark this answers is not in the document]
kronion: @araffin would you accept this black config change as a separate PR?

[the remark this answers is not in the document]
araffin: I will need to check if it keeps working on my machine, otherwise we can keep it there ;)

sb3_contrib/ppo_mask/ppo_mask.py
araffin: if the test env does not support it, it can still provide an invalid action method, no? (which will return a vector that says every action is valid)

sb3_contrib/common/maskable/utils.py
kronion: Thanks for the feedback. I've simplified the API, there's no longer an expectation that envs will have an `action_mask_fn` attribute.

sb3_contrib/common/wrappers/action_masker.py
```diff
@@ -0,0 +1,36 @@
...
+            found_method = getattr(self.env, action_mask_fn)
+            if not callable(found_method):
+                raise ValueError(f"Environment attribute {action_mask_fn} is not a method")
+
+            self.action_mask_fn = found_method
+        else:
+            self.action_mask_fn = action_mask_fn
+
+    def action_masks(self) -> np.ndarray:
+        return self.action_mask_fn(self.env)
```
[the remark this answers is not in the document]
kronion: It's a function that takes an env and returns the mask, it's not a bound env method. I've simplified the API so envs are only expected to expose an `action_masks()` method. Let me know if it's still unclear.

sb3_contrib/ppo_mask/ppo_mask.py
kronion: I refactored so that `use_masking` is a kwarg for the `learn()` method, rather than a model attribute specified at construction time. That way `use_masking` isn't included in the saved model.

It's now an error if an env (either training or eval) fails to implement the masking API when `use_masking` is true.

kronion:
@araffin @Miffyli I have a question about how best to support `MultiBinary` action spaces.

I thought each binary option in a `MultiBinary` would be represented by two logits, one for the 0 outcome and one for 1. That would have made it equivalent to a `Categorical` with two options. I was wrong about that, there's only one logit, associated with the 1 outcome.

The masking implementation for `Categorical` and `MultiCategorical` expects one boolean mask per logit. If I follow the same pattern for `MultiBinary`, that would mean it would be possible to mask out the 1 outcome, but not the 0 outcome. So a `MultiBinary` space would not be equivalent to a `Categorical` space with two options.

Personally I think this is reasonable. The binary in a `MultiBinary` is like a button, and it may be invalid to click the button, but it should never be invalid _not_ to click it. If that's really what you wanted, you should use a `Categorical`.

Let me know what you think. There are two other options I considered:
1. Allow the action mask to include two booleans per binary, one for each outcome. Relative to the shape of the action space, the shape of the action mask would have one more dimension of size 2.
2. Don't support `MultiBinary` at this time. Personally I'd like to include it, but I think it should be the last feature added to this PR.

Miffyli:
I am not sure what the confusion here is. MultiBinary is not similar to Categorical to any degree. It is just a special case of MultiCategorical. To me having an invalid-action mask for it seems completely reasonable (instead of 0/1 decision for one action, there is the third one that indicates "invalid action").

I say the simplest and least-surprising option here is to support MultiBinary in similar way as MultiCategorical is supported. This avoids odd, unexpected behavior when switching between the two.

vwxyzjn:
I agree with @Miffyli. MultiBinary is a special case of MultiDiscrete and therefore should be treated the same 🙂

kronion:
I should have been more exact, I was trying to draw a parallel between each individual binary in a `MultiBinary` and each `Categorical` in a `MultiCategorical`.

Here's an example to clarify the question:
```python
import torch as th
from sb3_contrib.common.distributions import MaskableMultiCategoricalDistribution

# MultiCategorical contains just one categorical with two outcomes
mcd = MaskableMultiCategoricalDistribution([2])
mcd.proba_distribution(th.Tensor([[0, 0]]))  # Both outcomes equal probability
len(mcd.distributions)  # Returns 1

# Note the two logits, one per outcome of the categorical
mcd.distributions[0].logits  # Returns tensor([[-0.6931, -0.6931]])
mcd.sample()  # Can return tensor([[0]]) or tensor([[1]])
mcd.apply_masking([True, False])
mcd.sample()  # Can only return tensor([[0]])
mcd.apply_masking([False, True])
mcd.sample()  # Can only return tensor([[1]])


# Note this class doesn't actually exist yet. sb3 has a BernoulliDistribution class used for
# MultiBinary spaces, so imagine that's been adapted.
from sb3_contrib.common.distributions import MaskableBernoulliDistribution

# MultiBinary with one binary with two outcomes
mbd = MaskableBernoulliDistribution(1)
mbd.proba_distribution(th.Tensor([[0]]))  # Note only one logit provided, vs two for the multicategorical dist

# Only one logit returned, corresponding to the "1" outcome
mbd.distribution.logits  # Returns tensor([[0.]])
mbd.sample()  # # Can return tensor([[0]]) or tensor([[1]])

# Should mask include a bool per binary outcome, to be consistent with the bool per categorical outcome above?
mbd.apply_masking([True, False])
mbd.sample()  # Can only return tensor([[0]])
mbd.apply_masking([False, True])
mbd.sample()  # Can only return tensor([[1]])
```

madScientist10:
Hello,

I assume this feature does not support Box action spaces, correct? Are there any plans for it? Or, can someone provide me any general instructions on how to implement it myself?

Thanks.

araffin:
>I assume this feature does not support Box action spaces, correct? Are there any plans for it? Or, can someone provide me any general instructions on how to implement it myself?

Invalid actions for Box environment is not really possible. For instance if you command robot position (3 degrees of freedom, x y z positions), what does that mean to mask x dimension? (because you still need to send a desired position to your robot).

madScientist10:
It simply means that the robot will have a constant position in the x axis and only move in the y, z axes. I don't understand why it can't make sense. Perhaps if you allow yourself to think of an example that sounds more reasonable to you, you can see that it has quite a purpose of realization. In fact, I wouldn't ask if I did not have such a problem myself.

araffin:
>It simply means that the robot will have a constant position in the x axis and only move in the y, z axes. 

then you can do that directly in the environment, no?
The issue is more from the practical implementation, it is probably an open questions.
With discrete actions, you can simply not sample the action that is invalid, with continuous actions, you predict mean and std for each dimension, you cannot really mask an action (or only inside the env).

madScientist10:
You cannot do that directly from the environment, since in that case the agent will have to learn by itself how to pick the valid set of actions each time. This is a lot more different.

For instance, in my case I have an action space that consists of 3 (or more) other concatenated action spaces (Box-type). Each time, the agent should perform actions dictated by one of the three action spaces, leaving the other 2 untouched (and, before you ask, they are set to have constant values). That's what masking in this case would do: force the agent to play with only a subset of the original action space, instead of allowing it to fiddle with everything.

In other words, what I am really thinking of is a way to shorten the length of the original action space vector dynamically during training. This is equivalent to masking invalid actions.

araffin:
>Each time, the agent should perform actions dictated by one of the three action spaces, leaving the other 2 untouched

What you are describing sounds a bit like [Reinforcement Learning with Parameterized Actions](https://arxiv.org/abs/1509.01644).
Again, it sounds like an open research problem with no clear solution (at least I did not see any paper yet solving what you are describing).

>You cannot do that directly from the environment, since in that case the agent will have to learn by itself how to pick the valid set of actions each time,

This reminds me of "input mapping" from [Shared Control Templates for Assistive Robotics](https://elib.dlr.de/135803/).
And you can do that inside the env, even if it won't be as efficient as providing the agent with the correct set of actions.

>In other words, what I am really thinking of is a way to shorten the length of the original action space vector dynamically during training. This is equivalent to masking invalid actions.

Although the idea is simple to explain, I don't think there is a straightforward way to implement that (unless you know how to do it?).

madScientist10:
Thank you for the suggestions. However, they are not suitable for my problem.

I'm not really sure how it could be implemented, that's why I ask. I'm not very familiar with the technical details of stable-baselines. If I were to give a naive suggestion for implementation, I would say to work with the sampling distribution within the model: instead of grabbing the action corresponding to the policy with the highest probability from the distribution, skip all the actions/policies that include performing an illegal action, and go with the first one that doesn't.

Miffyli:
@kronion 

Re MultiBinary: Aaah right, I see the problem :). The behaviour should be same as MultiCategorical's. I think it might even be beneficial to change behaviour of `BernoulliDistribution` so that it is just a `MultiCategoricalDistribution` where all actions have only two outcomes (e.g. creates a `MultiCategoricalDistribution` with correct parameters and just passes all function calls forward to it). This would simplify the code and solve this issue easily.

@araffin any objections against making BernoulliDistribution just to be a special case of MultiCategorical, instead of using Pytorch bernoulli distribution? This should not change outputs of functions etc.

araffin:
>@araffin any objections against making BernoulliDistribution just to be a special case of MultiCategorical, instead of using Pytorch bernoulli distribution? This should not change outputs of functions etc.

I think it's fine (unless it makes the code too complicated).

btw @kronion , dou you need help for documentation and tests? (I can help a bit in the coming weeks)

kronion:
@araffin Apologies for the slow response. I haven't had the time to focus on this PR for the past few weeks, but I expect to return to it in July (and hopefully get it over the finish line).

I'd definitely welcome help on documentation and tests. I've started on the docs locally, but it's a bit slow going because I'm unfamiliar with Sphinx. I haven't started on testing yet.

araffin:
>I'd definitely welcome help on documentation and tests. I've started on the docs locally, but it's a bit slow going because I'm unfamiliar with Sphinx. I haven't started on testing yet.

I've added basic testing and basic doc (please update it with your local doc ;)).
But more testing is required (to test all type of policies/distributions).

With dict obs support we should be able to more easily reproduce the results from the paper, no?

kronion:
Finally getting some time to return to this PR. I've added basic testing of the underlying distribution masking, ie `MaskableCategorical`. I plan to work on adding `MultiBinary` action space support next, which should hopefully make the PR feature-complete.

> With dict obs support we should be able to more easily reproduce the results from the paper, no?
Possibly it will make things simpler. When I last tried to reproduce the paper results, I started by rewriting the control (ie non-masking) model to use SB3's PPO instead of its own [custom implementation](https://github.com/vwxyzjn/invalid-action-masking/blob/master/invalid_action_masking/ppo_no_mask_4x4.py). But the result performed worse than the original paper's. So I never even tried to apply masking because my basis of comparison was different from the start.

How about I make a repo for reproducing the experiment, and we debug there?

araffin:
> How about I make a repo for reproducing the experiment, and we debug there?

how custom should that be?
i would prefer starting from a fork of the rl zoo (to reduce potential mistakes).

Apart from the custom network, the ppo implementation looks quite close to SB3 one, probably @vwxyzjn and I can help you making baselines results match ;)

kronion:
I'm happy to contribute a model to the zoo. In the meantime, I'm sharing [what I have so far](https://github.com/kronion/microrts-ppo-comparison) in the hopes that some forward progress can be made. This repo contains code to run both the original model and a new implementation using SB3's PPO. Currently the original model trains significantly faster than the SB3 version, and I'm still not sure why.

@vwxyzjn if you have any insight here, it would be greatly appreciated. I noticed you recently added a new PPO implementation to the reference repo, based on `openai/baselines`. I'm curious if you've tried using SB3's PPO as well.

araffin:
> I'm happy to contribute a model to the zoo. In the meantime, I'm sharing [what I have so far](https://github.com/kronion/microrts-ppo-comparison) in the hopes that some forward progress can be made. This repo contains code to run both the original model and a new implementation using SB3's PPO. Currently the original model trains significantly faster than the SB3 version, and I'm still not sure why.
> 
> @vwxyzjn if you have any insight here, it would be greatly appreciated. I noticed you recently added a new PPO implementation to the reference repo, based on `openai/baselines`. I'm curious if you've tried using SB3's PPO as well.

the actor/critic extractor/networks should be independent (but they are currently shared in your sb3 code)

kronion:
Thank you, I didn't realize. I don't see an easy way to have separate extractors using the common `ActorCriticPolicy`, so I'm guessing I'll have to make my own policy. Let me know if there's a better way.

(But please enjoy your vacation!)

vwxyzjn:
@kronion Hey thanks for all these awesome work! Sorry didn't input more earlier. 

> I noticed you recently added a new PPO implementation to the reference repo, based on openai/baselines.

That is correct. To my understanding, this [PPO implementation](https://github.com/vwxyzjn/invalid-action-masking/blob/54bfb37b939e8f9e77dcf96f79b7df4953e012f2/ppo.py) (without the invalid action masking part) would be identical to SB3's PPO implementation per [my blog posts](https://costa.sh/blog-the-32-implementation-details-of-ppo.html)

I am cloning your repo to check it out :)

vwxyzjn:
Couple quick comments:

1. The biggest issue is probably your `n_epochs=80`. This is going to blow up learning. Try instead with `n_epochs=4`.
2. Try with a different `num_envs` and `n_steps` like [here](https://github.com/vwxyzjn/invalid-action-masking/blob/54bfb37b939e8f9e77dcf96f79b7df4953e012f2/ppo.py#L50-L52).
2. Remove the `Max2d` stuff [here] and use neural network [here](https://github.com/vwxyzjn/invalid-action-masking/blob/54bfb37b939e8f9e77dcf96f79b7df4953e012f2/ppo.py#L214-L221) instead.

vwxyzjn:
Hey I did a little tuning as shown in my PR (kronion/microrts-ppo-comparison#1). Now you can train an SB3 model that gets much more rewards. See https://wandb.ai/costa-huang/sb3/runs/2gltlpfv?workspace=user-costa-huang or 

image

vwxyzjn:
That said, it's interesting... I'd expect it to get similar returns to https://github.com/kronion/microrts-ppo-comparison/blob/266856e528a3da258bad85fe74f30d14bed20036/original/new_train_ppo_4x4.py, which got 40. Hmmm my suspicion is the neural network configuration.

vwxyzjn:
This is what my agent looked like
```
Agent(
  (network): Sequential(
    (0): Conv2d(27, 16, kernel_size=(3, 3), stride=(2, 2))
    (1): ReLU()
    (2): Flatten(start_dim=1, end_dim=-1)
    (3): Linear(in_features=16, out_features=128, bias=True)
    (4): ReLU()
  )
  (actor): Linear(in_features=128, out_features=61, bias=True)
  (critic): Linear(in_features=128, out_features=1, bias=True)
)
```
This is what the SB3 agent looked in my PR
```
ActorCriticPolicy(
  (features_extractor): MicroRTSExtractor(
    (cnn): Sequential(
      (0): Conv2d(27, 16, kernel_size=(2, 2), stride=(1, 1))
      (1): ReLU()
      (2): Flatten(start_dim=1, end_dim=-1)
    )
  )
  (mlp_extractor): MlpExtractor(
    (shared_net): Sequential(
      (0): Linear(in_features=144, out_features=128, bias=True)
      (1): ReLU()
    )
    (policy_net): Sequential()
    (value_net): Sequential()
  )
  (action_net): Linear(in_features=128, out_features=61, bias=True)
  (value_net): Linear(in_features=128, out_features=1, bias=True)
)
```

vwxyzjn:
In the latest commit in the PR, the performance is already matched. 

https://wandb.ai/costa-huang/sb3/runs/3k2m2hlo?workspace=user-costa-huang

image

Hope this helps.

araffin:
>In the latest commit in the PR, the performance is already matched.

thanks for the support =)
What were the main differences?

@kronion once you have merged the PR and reproduced the results, please update the doc and we are probably good to go ;) (maybe only the `MultiBinary` doc/implementation is missing as there is an unused import in `distributions.py` for now)

vwxyzjn:
@araffin that’s is a couple: see here https://github.com/kronion/microrts-ppo-comparison/pull/1/files#diff-557e74960a5b99e54f2eb1a79355f7aaee007bf02211d44ae20a695a225b01d0.

the main things are the hyper parameters change, layer initialization, `observations.permute((0, 3, 1, 2))`

araffin:
>observations.permute((0, 3, 1, 2))

the env was channel last? normally SB3 re-order the channels automatically...

vwxyzjn:
@araffin hmm interesting. I think @kronion uses the custom networks and if you look at the logs at [here](https://wandb.ai/costa-huang/sb3/runs/3k2m2hlo/logs?workspace=user-costa-huang) (you need to click the download button on the top right) as shown below, the `VecTransposeImage` wasn't used. 

```
Logging to zoo/4x4/2021-08-17T16:47:21
/home/costa/microrts/maps/4x4/basesWorkers4x4.xml
[Lai.rewardfunction.RewardFunctionInterface;@2d554825
/home/costa/microrts/maps/4x4/basesWorkers4x4.xml
[Lai.rewardfunction.RewardFunctionInterface;@6be46e8f
/home/costa/microrts/maps/4x4/basesWorkers4x4.xml
[Lai.rewardfunction.RewardFunctionInterface;@3567135c
/home/costa/microrts/maps/4x4/basesWorkers4x4.xml
[Lai.rewardfunction.RewardFunctionInterface;@327471b5
/home/costa/microrts/maps/4x4/basesWorkers4x4.xml
[Lai.rewardfunction.RewardFunctionInterface;@4157f54e
/home/costa/microrts/maps/4x4/basesWorkers4x4.xml
[Lai.rewardfunction.RewardFunctionInterface;@90f6bfd
/home/costa/microrts/maps/4x4/basesWorkers4x4.xml
[Lai.rewardfunction.RewardFunctionInterface;@47f6473
/home/costa/microrts/maps/4x4/basesWorkers4x4.xml
[Lai.rewardfunction.RewardFunctionInterface;@15975490
Using cuda device
8
ActorCriticPolicy(
  (features_extractor): MicroRTSExtractor(
    (cnn): Sequential(
      (0): Conv2d(27, 16, kernel_size=(3, 3), stride=(2, 2))
      (1): ReLU()
      (2): Flatten(start_dim=1, end_dim=-1)
    )
  )
  (mlp_extractor): MlpExtractor(
    (shared_net): Sequential(
      (0): Linear(in_features=16, out_features=128, bias=True)
      (1): ReLU()
    )
    (policy_net): Sequential()
    (value_net): Sequential()
  )
  (action_net): Linear(in_features=128, out_features=61, bias=True)
  (value_net): Linear(in_features=128, out_features=1, bias=True)
)
Logging to runs/3k2m2hlo/PPO_1
```

araffin:
Looking at the observation space (https://github.com/vwxyzjn/gym-microrts/blob/master/gym_microrts/envs/vec_env.py#L94), that's normal, SB3 only consider uint8 + values in [0, 255] as image.
See https://github.com/DLR-RM/stable-baselines3/blob/3efab0d267e74cb03264411d4500ddde0c163404/stable_baselines3/common/preprocessing.py#L27

vwxyzjn:
> it's interesting because the networks are still shared in your PR (compared to your implementation)

Did you mean the value and policy head in my original implementation of invalid action masking does not share parameters? That is true and that’s why I am recommending people to use https://github.com/kronion/microrts-ppo-comparison/blob/266856e528a3da258bad85fe74f30d14bed20036/original/new_train_ppo_4x4.py instead of the original one.

kronion:
Hey everyone. Trying to catch up on all the updates here.

vwxyzjn, thanks for the help debugging model performance. `n_epochs=80` was definitely a major issue. I retested my version of the sb3 model adding in your changes one at a time, and that hyperparameter tweak appears to be the key change:

Screenshot from 2021-08-24 10-00-09

The other changes to `extractors.py` (cutting the `MaxPool2D` layer, changing the layer initialization) don't appear to affect the performance. Note that three of the runs are overlapping in the chart above.

I also made an attempt at having separate feature extractor layers for the policy and value networks. I did two versions of this, one where there was still a shared layer in the MLP after the feature extractors, and one where there were no shared layers at all. Neither version was an improvement over the original model after 100k timesteps. 

Screenshot from 2021-08-24 21-32-36

Ultimately I'm still unsure why it takes so many more timesteps for the SB3 model to converge than vwxyzjn's original model did, but it's possible that it's not necessary to get to the bottom of that. If we're willing to accept this version of the SB3-based model as a reasonable baseline, it should be straightforward to show that masking speeds up convergence.

vwxyzjn:
@kronion the `n_epochs=80` is indeed an issue. You would only use `n_epochs=80` with the early stopping optimization.

The following screenshot shows the performance in Halfcheetah-v2, where `n_epochs=80` with the early stopping optimization (purple line) is much better than just `n_epochs=80`.

image

See https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=9520424 for more details.

araffin:
Looking at the code you linked, why is there nothing with `MaskablePPO` in it?
(an easy way to make it like original PPO is to overwrite the invalid action mask and say that all actions are valid)

>Neither version was an improvement over the original model after 100k timesteps.

we are talking about PPO without any invalid masking, right?

araffin:
I've added a version with maskable PPO: https://github.com/kronion/microrts-ppo-comparison/pull/2
and the first run looks good: 
eval

kronion:
> Looking at the code you linked, why is there nothing with MaskablePPO in it?
> ...
> we are talking about PPO without any invalid masking, right?

Yes, I was only trying to reproduce the non-masked results of the original paper, which converged on a reward of 40 by 100k timesteps instead of ~250k like we're seeing using SB3's PPO.

At this point it seems like that level of reproducibility isn't needed, and it will suffice to show that MaskablePPO converges faster than SB3's PPO.

So it looks like we're very nearly at the finish line. @araffin looking at the checklist at the top:
- You've effectively reproduced the paper results. Is that ready to be checked?
- We have documentation and some tests. Please let me know if we need more of either.
- I'm not sure how you'd like to include baseline results. Does that mean moving the example models into the zoo?
- The rest is just cleanup :smile:

araffin:
yes, we are close to merging this PR =D!

> At this point it seems like that level of reproducibility isn't needed, and it will suffice to show that MaskablePPO converges faster than SB3's PPO.

Well, it would still be good to match the results (probably using the new script from @vwxyzjn ) as it is never a good sign, but we may leave it from later.

>You've effectively reproduced the paper results. Is that ready to be checked?

The script need to be adapted to allow to use masking and non-masking and please do at least 3 runs for 4x4 and 10x10
+ add the results to first message of that PR.
You also need to update the install instructions slightly (a symlink is needed so the microrts maps are found, I had some troubles).

>We have documentation and some tests. Please let me know if we need more of either.

currently missing: reproducing results documentation.

for the tests, I think I had left some TODOs at the bottom.
I also remember that the `ActionWrapper` is currently un-tested (you can check the coverage report to see what's missing too).

>I'm not sure how you'd like to include baseline results. Does that mean moving the example models into the zoo?

If it does require too much efforts, please do, otherwise, instructions in the doc on how to use your repo would be fine here (as the env is custom).

araffin:
While trying to investigate PPO performance with 10x10 map, I found a bug: the current implementation does not work with n_envs>1, is that intended?

The error:

```
sb3_contrib/ppo_mask/ppo_mask.py:515: in learn
    continue_training = self.collect_rollouts(self.env, callback, self.rollout_buffer, self.n_steps, use_masking)
sb3_contrib/ppo_mask/ppo_mask.py:342: in collect_rollouts
    action_masks=action_masks,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <sb3_contrib.common.maskable.buffers.MaskableRolloutBuffer object at 0x7f4ec1dfbfd0>
action_masks = array([[False, False, False,  True,  True,  True, False, False, False,
         True,  True,  True, False, False, Fals...lse,  True, False, False, False, False, False,  True,
        False, False,  True, False,  True,  True, False,  True]])
args = (array([18, 18]), array([[11],
       [33]]), array([0., 0.], dtype=float32), array([ True,  True]), tensor([[0.0068],
        [0.0068]]), tensor([-2.9933, -2.9953]))
kwargs = {}

    def add(self, *args, action_masks: Optional[np.ndarray] = None, **kwargs) -> None:
        """
        :param action_masks: Masks applied to constrain the choice of possible actions.
        """
        if action_masks is not None:
>           self.action_masks[self.pos] = action_masks.flatten()
E           ValueError: could not broadcast input array from shape (160) into shape (2,80)

sb3_contrib/common/maskable/buffers.py:54: ValueError
```

sb3_contrib/common/envs/invalid_actions_env.py
```diff
@@ -46,6 +46,10 @@ def __init__(
         self.num_resets = -1  # Becomes 0 after __init__ exits.
         self.reset()
 
+    def seed(self, seed: int) -> None:
+        self.action_space.seed(seed)
+        np.random.seed(seed)
+
```
kronion, line 52: Was this change intended?

vwxyzjn:
> EDIT: with the new script and the branch https://github.com/araffin/microrts-ppo-comparison/tree/feat/10x10, I could match the new results ;)

Yep yep, @kronion and @araffin, since the original implementation from the paper did not match the implementation details from openai/baselines, I would suggest matching the performance of the new script instead.

kronion:
Thanks @vwxyzjn. One other question for you. We're still running with `gym-microrts==0.1.3`. That's a very old version at this point, but from 0.1.4 onward the env IDs used in the paper are missing. `MicrortsMining10x10F9-v0`, etc. I don't think it's an issue, but if there's a way to update to a newer version of that library, it wasn't obvious to me.

vwxyzjn:
@kronion I have thought about making the experiments more reproducible, hence have added the reproducibility info here: https://github.com/vwxyzjn/invalid-action-masking#get-started. 

Unfortunately, it's rather difficult to bring back old stuff to the latest release. I thinking going forward, my best resolution for this is to support the PPO variant in this branch with the latest gym-microrts structure.

kronion:
@araffin 
> The script need to be adapted to allow to use masking and non-masking and please do at least 3 runs for 4x4 and 10x10 
> \+ add the results to first message of that PR.
> You also need to update the install instructions slightly (a symlink is needed so the microrts maps are found, I had some troubles).

I've done all of that in https://github.com/kronion/microrts-ppo-comparison:
- README updated to clarify how to install microrts
- SB3 script supports masking and non-masking
- SB3 script requires environment size to be specified as a parameter (either 4 or 10)
- Completed 3 runs for each tuple of (environment size, masking?). Checked in results as tensorboard logs.
- Added summary of the results as screenshots in the README.

Please let me know if those results should be copied into this PR, either as comments in this conversation chain or in some other form.

araffin:
> I've done all of that in https://github.com/kronion/microrts-ppo-comparison:

perfect =)
I updated the code to reflect the changes that were made in SB3 (train/eval mode, better estimator for kl div, ...)

>Please let me know if those results should be copied into this PR, either as comments in this conversation chain or in some other form.

I added a link in the PR ;)
(but the documentation need to be updated too)

vwxyzjn:
Thanks for this great progress. Recently I have made a change that should make the latest gym-microrts somewhat compatible. Similar to how the [chess's action space](https://www.pettingzoo.ml/classic/checkers) is defined in pettingzoo, in a 16x16 map the gym-microrts' action space is now a 16x16x78 dimensional array. To make it compatible with gym, the action space now is [`MultiDiscrete(np.array([[6, 4, 4, 4, 4, len(self.utt['unitTypes']), 7 * 7]] * 16 * 16).flatten())`](https://github.com/vwxyzjn/gym-microrts/blob/769b6a294269a72bb967333486c31a2fc348f2e5/gym_microrts/envs/new_vec_env.py#L100)

I have started a repo to try out the `MaskablePPO`: https://github.com/vwxyzjn/microrts-sb3, but have encountered an issue. The current setup allows the users to set a mask function at the gym environment level, but not the vectorized environment level, so this make it difficult to work... In `gym-microrts>=0.4.3`, there is a `envs.get_action_mask` function that returns the mask for the vector environment, so not sure what's the best way to move forward.

araffin:
>  Recently I have made a change that should make the latest gym-microrts somewhat compatible. Similar to how the chess's action space is defined in pettingzoo,

@vwxyzjn  I would suggest to tackle that in a separate issue?
Same for the dict obs support.
I would first merge a (not so) minimal working version before adding more features ;)

araffin (approved):
LGTM, thanks a lot for all your efforts =)

jeweinb:
I have a question regarding issue #580. Does this action masking architecture support a more dynamic masking of actions. Such as given current state what are the available actions for the agent to select... A(s). I see that this supports a masked action set based on environment. But it is not clear if it supports an action set based on state.

Miffyli:
@vahidqo The error suggests the action and mask dimensionality does not match. Double-check that your mask has as many values as there are possible actions.

vahidqo:
@Miffyli thanks for your answer. my action space is:
`self.action_space = MultiDiscrete([self.action_dim1,self.action_dim2])`
and my mask is :
```
    def _update_mask(self, state):
        action_mask = np.zeros((2,max(self.action_dim1,self.action_dim2)), dtype=bool)
        # Wait and return to restaurant are always allowed
        action_mask[1][0:self.action_dim2] = True
        action_mask[0][0] = True
        action_mask[0][0:min(int(self.bike[int(self.vehicle_location[0])]),self.vehicle_capacity-int(self.vehicle_load[0]))] = True
        action_mask[0][self.vehicle_capacity+1:self.vehicle_capacity+int(self.vehicle_load[0])] = True
        

        return action_mask
```
and my mask_fn is:
```
def mask_fn(env: gym.Env) -> np.ndarray:

    return env._update_mask(env.state)

```

could you please help me?

rllyryan:
For the code below:

```
import gym
import numpy as np

from sb3_contrib.common.maskable.policies import MaskableActorCriticPolicy
from sb3_contrib.common.wrappers import ActionMasker
from sb3_contrib.ppo_mask import MaskablePPO


def mask_fn(env: gym.Env) -> np.ndarray:
    # Do whatever you'd like in this function to return the action mask
    # for the current env. In this example, we assume the env has a
    # helpful method we can rely on.
    return env.valid_action_mask()


env = ...  # Initialize env
env = ActionMasker(env, mask_fn)  # Wrap to enable masking

# MaskablePPO behaves the same as SB3's PPO unless the env is wrapped
# with ActionMasker. If the wrapper is detected, the masks are automatically
# retrieved and used when learning. Note that MaskablePPO does not accept
# a new action_mask_fn kwarg, as it did in an earlier draft.
model = MaskablePPO(MaskableActorCriticPolicy, env, verbose=1)
model.learn()

# Note that use of masks is manual and optional outside of learning,
# so masking can be "removed" at testing time
model.predict(observation, action_masks=valid_action_array)
```

**Is the mask made of a numpy array of zeros and ones?**

araffin:
> Is the mask made of a numpy array of zeros and ones?

yes, actually an array of boolean, you can find an example here: https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/blob/master/sb3_contrib/common/envs/invalid_actions_env.py#L35-L36

rllyryan:
> > Is the mask made of a numpy array of zeros and ones?
> 
> yes, actually an array of boolean, you can find an example here: https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/blob/master/sb3_contrib/common/envs/invalid_actions_env.py#L35-L36

Thank you for your reply, can I ask if the minimal usage example at the very top of the page still applies? It seems useful for environments that have different action masks depending on the state of the env.

pmarantis:
I see that a few people are looking for implementation of maskable PPO on custom envs, so in this fun project:   https://github.com/pmarantis/2048-game-gym-custom-Env-for-maskable-PPO is my use of maskable PPO on a custom env for the game 2048. Any review on my attempt and if the implementation is applied correctly, as well as any tips on improving the agent results (like a better reward function or using cnnpolicy) is appreciated. Possibly this could be used as an example for maskable ppo on a custom env and how the masking is done. Thanks to the team of sb for this interesting variant of PPO !

Ranhuiryan:
After several frustrating attempts, I found that this implementation is still not compatible with the latest gymnasium API. Can someone update it?

PR #35, opened 2021-07-02
Add the Bootstrapped Dual Policy Iteration algorithm for discrete action spaces

steckdenis:
This sb3_contrib pull request follows [an issue opened in stable-baselines3](https://github.com/DLR-RM/stable-baselines3/issues/499).

## Description

This pull request adds the Bootstrapped Dual Policy Iteration algorithm to stable-baselines3-contrib, with documentation and updated unit tests (I was able to make them work by replacing logger.record, as in SB3 algos, with self.logger.record, as in the other SB3-contrib algos).

The original BDPI paper is https://arxiv.org/abs/1903.04193. The main reason I propose to have BDPI in stable-baselines3-contrib is that it is quite different from other algorithms, as it heavily focuses on sample-efficiency at the cost of compute-efficiency (which is nice for slow-sampled robotic tasks). The main results in the paper show that BDPI outperforms several state-of-the-art RL algorithms on many environments (Table is difficult to explore into, and Hallway comes from gym-miniworld and is a 3D environment):

bdpi_results

I have reproduced these results with the BDPI algorithm I propose in this PR, on LunarLander. PPO, DQN and A2C have been run using the default hyper-parameters used by rl-baselines3-zoo (I suppose the tuned ones?), for 8 random seeds each. The BDPI curve is also the result of 8 random seeds. I apologize for the truncated runs, BDPI and DQN only ran for 100K time-steps (for BDPI, due to time constraints, as it takes about an hour per run to perform 100K time-steps):

bdpi_stablebaselines3

## Types of changes

- [ ] Bug fix: none
- [x] New feature: Policies for the BDPI actor and critics, with documentation. Supported policies are MlpPolicy, CnnPolicy and MultiInputPolicy.
- [x] New feature: The BDPI algorithm, with a simple (non-invasive in the code) multi-processing approach, to help with the compute requirements. I aimed at writing the code in the easiest way possible to understand.
- [ ] Breaking change: none
- [x] Tests: The unit tests have been updated for BDPI, to the best of my knowledge. Every test with BDPI passes, there is no regression.
- [x] Documentation: I have updated the main page, TOC and Examples. I have written a page dedicated to the BDPI algorithm. Once compiled, the documentation looks nice on my machine, but I'm quite a novice Sphinx user.
- [x] My changes to rl-baselines3-zoo (hyper-parameter definition, and tuned hyper-parameters for LunarLander) are here: https://github.com/steckdenis/rl-baselines3-zoo

## Checklist:

- [ ] I have included baseline results: I'm not sure if my plot above counts as baseline results, or if there is something more specific to do?
- [ ] I have updated the changelog accordingly: There is no alpha version in the changelog that I can easily update (only Release 1.1.0 that is finalized). I would need help regarding the creation of a new alpha version after 1.1.0

### DIFF
--- a/docs/guide/algos.rst
+++ b/docs/guide/algos.rst
@@ -10,11 +10,12 @@ Name         ``Box``     ``Discrete`` ``MultiDiscrete`` ``MultiBinary`` Multi Pr
 ============ =========== ============ ================= =============== ================
 TQC          ✔️          ❌            ❌                ❌              ❌
 QR-DQN       ️❌          ️✔️            ❌                ❌              ❌
+BDPI         ❌          ✔️            ❌                ❌              ✔️
 ============ =========== ============ ================= =============== ================
 
 
 .. note::
-    Non-array spaces such as ``Dict`` or ``Tuple`` are not currently supported by any algorithm.
+    Non-array spaces such as ``Dict`` or ``Tuple`` are only supported by BDPI, using ``MultiInputPolicy`` instead of ``MlpPolicy`` as the policy.
 
 Actions ``gym.spaces``:
 
--- a/docs/modules/bdpi.rst
+++ b/docs/modules/bdpi.rst
@@ -1,0 +1,136 @@
+.. _bdpi:
+
+.. automodule:: sb3_contrib.bdpi
+
+BDPI
+====
+
+`Bootstrapped Dual Policy Iteration <https://arxiv.org/abs/1903.04193>`_ is an actor-critic algorithm for
+discrete action spaces. The distinctive components of BDPI are as follows:
+
+- Like Bootstrapped DQN, it uses several critics, with each critic having a Qa and Qb network
+  (like Clipped DQN).
+- The BDPI critics, inspired from the DQN literature, are therefore off-policy. They don't know
+  about the actor, and do not use any form of off-policy corrections to evaluate the actor. They
+  instead directly approximate Q*, the optimal value function.
+- The actor is trained with an equation inspired from Conservative Policy Iteration, instead of
+  Policy Gradient (as used by A2C, PPO, SAC, DDPG, etc). This use of Conservative Policy Iteration
+  is what allows the BDPI actor to be compatible with off-policy critics.
+
+As a result, BDPI can be configured to be highly sample-efficient, at the cost of compute efficiency.
+The off-policy critics can learn aggressively (many samples, many gradient steps), as they don't have
+to remain close to the actor. The actor then learns from a mixture of high-quality critics, leading
+to good exploration even in challenging environments (see the Table environment described in the paper
+linked above).
+
+Can I use?
+----------
+
+-  Recurrent policies: ❌
+-  Multi processing: ✔️
+-  Gym spaces:
+
+
+============= ====== ===========
+Space         Action Observation
+============= ====== ===========
+Discrete      ✔️      ✔️
+Box           ❌     ✔️
+MultiDiscrete ❌     ✔️
+MultiBinary   ❌     ✔️
+Dict          ❌     ✔️
+============= ====== ===========
+
+Example
+-------
+
+Train a BDPI agent on ``LunarLander-v2``, with hyper-parameters tuned by Optuna in rl-baselines3-zoo:
+
+.. code-block:: python
+
+  import gym
+
+  from sb3_contrib import BDPI
+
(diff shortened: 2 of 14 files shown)


docs/guide/algos.rst
```diff
@@ -10,11 +10,12 @@ Name         ``Box``     ``Discrete`` ``MultiDiscrete`` ``MultiBinary`` Multi Pr
 ============ =========== ============ ================= =============== ================
 TQC          ✔️          ❌            ❌                ❌              ❌
 QR-DQN       ️❌          ️✔️            ❌                ❌              ❌
+BDPI         ❌          ✔️            ❌                ❌              ✔️
 ============ =========== ============ ================= =============== ================
 
 
 .. note::
-    Non-array spaces such as ``Dict`` or ``Tuple`` are not currently supported by any algorithm.
```
araffin, line 17: it seems that we forgot to update the contrib doc when adding support for Dict obs, the correct formulation should be the one from https://github.com/DLR-RM/stable-baselines3/blob/master/docs/guide/algos.rst

docs/guide/algos.rst
```diff
@@ -10,11 +10,12 @@ Name         ``Box``     ``Discrete`` ``MultiDiscrete`` ``MultiBinary`` Multi Pr
 ============ =========== ============ ================= =============== ================
 TQC          ✔️          ❌            ❌                ❌              ❌
 QR-DQN       ️❌          ️✔️            ❌                ❌              ❌
+BDPI         ❌          ✔️            ❌                ❌              ✔️
```
araffin, line 13: it says "multiprocessing" but I only see tests with one environment in the code...
and you probably need https://github.com/DLR-RM/stable-baselines3/pull/439 to make it work with multiple envs.

docs/modules/bdpi.rst
```diff
@@ -0,0 +1,136 @@
...
+      env.render()
+
+
+Results
+-------
+
+LunarLander
+^^^^^^^^^^^
+
+Results for BDPI are available in `this Github issue <https://github.com/DLR-RM/stable-baselines3/issues/499>`_.
```
araffin, line 89: as you are aiming for sample efficiency, I would prefer a comparison to DQN, QR-DQN (with tuned hyperparameters, results are already linked in the documentation: https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/pull/13)

Regarding which envs to compare too, please do the classic control ones + 2 Atari games at least (Pong, Breakout) using the zoo, so we can compare the results with QR-DQN and DQN.

araffin: I would also like a comparison of the compromise sample efficiency vs training time (how much more time does it take to train?)

sb3_contrib/bdpi/bdpi.py
```diff
@@ -0,0 +1,281 @@
...
+        random.shuffle(critics)
+
+        for criticA, criticB in critics:
+            # Sample replay buffer
+            with th.no_grad():
+                replay_data = self.replay_buffer.sample(batch_size, env=self._vec_normalize_env)
+
+            # Update the critic (code taken from DQN)
+            with th.no_grad():
+                qvA = criticA(replay_data.next_observations)
```
araffin, line 209: please use meaningful variable names

docs/guide/algos.rst
steckdenis: I may have mis-understood what "Multiprocessing" in this document means, and what PPO in stable-baselines is doing.

BDPI distributes its training updates on several processes, even if only one environment is used. To me, this is multi-processing, comparable to PPO that uses MPI to distribute compute. But if PPO uses multiple environments to be able to do multiprocessing, then I understand that "Multiprocessing" in the documentation means "compatible with multiple envs", not just "fast because it uses several processes".

Should I add a note, or a second column to distinguish "multiple envs" from "multi-processing with one env"?

araffin: >To me, this is multi-processing, comparable to PPO that uses MPI to distribute compute. 

PPO with MPI distributes env and training compute (and is currently not implemented in SB3).

>then I understand that "Multiprocessing" in the documentation means "compatible with multiple envs"

yes, that's the meaning (because we can use `SubProcEnv` to distribute data collection)

>Should I add a note, or a second column to distinguish "multiple envs" from "multi-processing with one env"?

no, I think we have already enough columns in this table.

Issue #11, opened 2020-11-26
Implement PPO MPI (SB2 PPO1)


araffin:
MPI can be quite useful to use multiprocessing [full potential](https://twitter.com/hardmaru/status/1260852988475658242)
but is dependency that can be tricky to install.

vwxyzjn:
Hey @araffin I prototyped multi-GPU support with `torch.distributed` https://github.com/vwxyzjn/cleanrl/pull/162. Preliminary experiments seem successful when controlling torch thread number to 1 per process and use SyncVecEnv:

`ppo_atari_multigpu_batch_reduce.py` was able to obtain 20-30% speed up at no cost of efficiency by leveraging data parallelism given by  `torch.distributed`.

![image](https://user-images.githubusercontent.com/5555347/164087727-ce7481a1-823d-4ded-9fd1-788eca7b84f4.png)

araffin:
@vwxyzjn ooh nice =)
I didn't know you could do that with PyTorch (it looks like they included MPI but for GPU).
I will try to have a look later this week.
Does this work also with cpu only?

PR #42, opened 2021-09-22
Augmented Random Search (ARS)

### LINKED ISSUE #41: Contributing ARS implementation.
Moving https://github.com/DLR-RM/stable-baselines3/issues/565 to here per request. 

### 🚀 Feature

I would like to contribute an [ARS ](https://arxiv.org/pdf/1803.07055.pdf) implementation to SB3.

### Motivation

BRS/ARS is a dead simple algorithm that works more often than most people would expect. I wish more people would try it on their custom environments (and maybe realize they don't really need DRL at all, or that their environment is too easy etc). 

### Pitch

I contribute a version of ARS that is compatible with and to up to the standards of SB3.

I have a version I made for myself [here](https://github.com/sgillen/seagul/blob/master/seagul/rl/ars/ars_torch.py), and there are [other implementations](https://github.com/ray-project/ray/blob/master/rllib/agents/ars/ars.py) I can compare performance to. ARS is very simple, though some care will be needed to find the most natural / least intrusive way to incorporate it into the existing code base. If you are down with the general idea I can come up with a proposal on that front as well. 

I'd also like to run and contribute ARS agents to the sb3-zoo.

I'm happy to do all this with minimal help, but only if the team actually thinks this is useful and will merge it. 

### Alternatives

1. <del> Add this to the sb3-contrib repo instead. I don't think ARS is super cutting edge or fancy, but possibly the team would prefer I put it there anyway.</del>

2. I don't write this algorithm at all and continue using my own version and grumbling that people aren't benchmarking against this simple baseline. 

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/DLR-RM/stable-baselines3/issues) in the repo (**required**)

sgillen:
## Description
closes #41

WIP pull request for ARS. Still some work to be done, but ready for some feedback before moving forward.

Original paper: https://arxiv.org/abs/1803.07055

A few questions that I would like feedback on
- I don't plan to support CNN policies for ARS. It is certainly possibly to make ARS support them, but not really what ARS was meant for. Do you think that is alright for now?
- I only support Box spaces for action and observation, the only other one that I think makes sense to add is dict observations, which I will probably add. But if other spaces should be included let me know. 
- I validate user hyper-parameters when learn is called, and will automatically adjust them if n_delta is not a multiple of n_workers. I could support the case where this condition is not met but it will considerably complicate the code to support a use case that doesn't really provide a benefit and makes the training slower.  

Critical feedback on the code quality is definitely welcome, I'm not super happy with the code right now, it seems more complicated than it needs to be, but I think its what makes the most sense for sb3 vec envs. Also, I put a lot of time trying and tried several approaches to make ARS fast with large env# subprocvecenvs but ultimately I find it's considerably faster to run single threaded (1 env dummyvec) and run multiple seeds in parallel.   Perhaps in the future I can make an MPI version along with [PPO](https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/issues/11)

## Context

- [x] I have raised this [issue](https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/issues/41) to propose this change ([required](https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/blob/master/CONTRIBUTING.md))

## Types of changes

## Checklist:

- [x] I have included baseline results (**required** for new training algorithms or training-related features).

## Baseline Reward Curves

The original paper uses v1 versions of the Mujoco envs. getting back to the old V1 envs will require an ancient version of gym and mujoco_py, which IIRC have breaking changes and are not compatible with sb3. But I did train on v2 using hyper params from original paper, using [my fork](https://github.com/sgillen/rl-baselines3-zoo/tree/feat/ars) of zoo. Each environment trained with 8 random seeds. Results are marginally worse than the paper, but seem fine to me given the environment version difference and the fact that only 3 seeds are presented for the optimized params in the paper. However If you'd like it should be possible to compare the papers original implementation on the newer envs. 

I can also train on v2 for Walker, Ant, and Humanoid, they just take a bit longer, and I'm not sure what the differences between v1 and v2 are there. However perhaps instead it might make sense to ignore Mujoco and to run optuna on the bullet envs and box2d before merging, since I want to do this in any case. I have several workstations to put to this task, but it will still take quite some time.

Results_Swimmer
Results_Hopper
Results_HalfCheetah


Miffyli:
> I don't plan to support CNN policies for ARS. It is certainly possibly to make ARS support them, but not really what ARS was meant for. Do you think that is alright for now?

Yup this is alright. Original paper did not have CNN policies, and I think it would require some more tweaking to get them working in reasonable time.

> I only support Box spaces for action and observation, the only other one that I think makes sense to add is dict observations, which I will probably add. But if other spaces should be included let me know.

This is alright for now, too. Anything beyond the original implementation is "nice to have but not necessary". I would recommend using your energy trying to match the original results as closely as humanly possible, rather than adding new things on top :)

> I validate user hyper-parameters when learn is called, and will automatically adjust them if n_delta is not a multiple of n_workers. I could support the case where this condition is not met but it will considerably complicate the code to support a use case that doesn't really provide a benefit and makes the training slower.

Sounds ok as it is, as long there is a warning message about this (ps: have not checked the code yet).

> Results are marginally worse than the paper, but seem fine to me given the environment version difference and the fact that only 3 seeds are presented for the optimized params in the paper. However If you'd like it should be possible to compare the papers original implementation on the newer envs.

That'd be nice. A simple table of the final agent performances side-by-side (with the variances included) would be the best case. We can then determine if it is close enough or something needs tweaking. Again, repeating experiments on the environments original paper used is enough (or on a subset of them).

sb3_contrib/ars/ars.py
```diff
@@ -0,0 +1,276 @@
...
+
+                candidate_returns[theta_idx - self.n_workers : theta_idx] = vec_returns
+                batch_steps += sum(vec_steps_taken)  # only count steps used in the update step towards our total_steps
+
+            callback.on_rollout_end()
+
+        return candidate_returns, batch_steps
+
+    # Make sure our hyper parameters are valid and auto correct them if they are not
+    def _validate_hypers(self):
```
Miffyli, line 183: Avoid unnecessary shorthands (change to `_validate_hyperparameters` or `_validate_hyperparams`).

sb3_contrib/ars/ars.py
```diff
@@ -0,0 +1,276 @@
...
+        self.n_params = len(self.theta)
+        self.policy = self.policy.to(self.device)
+
+    def _collect_rollouts(self, policy_deltas, callback):
+        with th.no_grad():
+            batch_steps = 0
+            theta_idx = 0
+
+            # Generate 2*n_delta candidate policies by adding noise to the current theta
+            candidate_thetas = np.concatenate([self.theta + policy_deltas, self.theta - policy_deltas])
```
Miffyli, line 130: Hmm would it not be more sensible to do all operations with torch tensors instead of creating numpy arrays and moving them back-and-worth torch and numpy?

sb3_contrib/ars/ars.py
```diff
@@ -0,0 +1,276 @@
...
+            callback.on_rollout_start()
+            while theta_idx < candidate_returns.shape[0]:
+                policy_list = []
+
+                # We are using a vecenv with n_envs==n_workers. We batch our candidate theta evaluations into vectors
+                # of length n_workers
+                for _ in range(self.n_workers):
+                    theta_tensor = th.tensor(candidate_thetas[theta_idx], dtype=self.dtype)
+                    th.nn.utils.vector_to_parameters(theta_tensor, self.policy.parameters())
+                    policy_list.append(copy.deepcopy(self.policy))
```
Miffyli, line 143: Hmm this is probably a big bottleneck. Wonder if you could create a population of N whose parameters you keep changing (or does population size change?)

araffin:
I won't have time (not energy) before next week to a full review, but just a quick remark:

>I only support Box spaces for action and observation

I would disagree with that as adding `Discrete` support should be fairly easy, see https://github.com/osigaud/stable-baselines3/blob/64baf2125dec791c20c41f4b39ed1dc2ad914010/stable_baselines3/cem/policies.py#L60

Also, you are not squashing the output for continuous actions (which means that they will be clipped automatically instead), is there a reason? (btw, if you do, you need to update call to parent class)


I saw that you overloaded `load` to default to cpu?
Why is that? And can't you do everything with torch? (removing the need of numpy)

and as overall remarks, same I did for TRPO, please use meaningful names and avoid mathematical notation when possible.

araffin:
Last remark, I know you are trying to have an efficient implementation by using `n_envs` and and the candidates in parallel, 
but I think it would be clearer (and less prone to errors) to use `evaluate_policy` as in the CEM implementation: https://github.com/osigaud/stable-baselines3/blob/64baf2125dec791c20c41f4b39ed1dc2ad914010/stable_baselines3/cem/cem.py#L156

sb3_contrib/ars/ars.py
sgillen: during profiling I saw that almost all the time is spent in env.step or policy.predict. ARS typically uses linear policies so the amount of data here is actually fairly small. preallocating the list would help a bit, perhaps just initializing with policy_list =[copy.deepcopy(policy) for _ in range(n_workers)], IIRC list comprehensions avoid the extra copies that append can run into.

sgillen:
> I would disagree with that as adding Discrete support should be fairly easy, see https://github.com/osigaud/stable-baselines3/blob/64baf2125dec791c20c41f4b39ed1dc2ad914010/stable_baselines3/cem/policies.py#L60

Will do

> Also, you are not squashing the output for continuous actions (which means that they will be clipped automatically instead), is there a reason? (btw, if you do, you need to update call to parent class)

No good reason, I think I was changing this while debugging performance and never changed it back. The original paper does not squash though, and I believe it may have a small effect on the learning performance. I can check if this is true. 

>I saw that you overloaded load to default to cpu?
Why is that? 

This was changed to pass some of the existing tests.I think ARS should default to CPU because it prefers linear policies or small MLPs. So when one creates a policy object it will default to CPU, but when loading will default to "auto" which searches for a CUDA device, this causes a test to fail and I think is not what the user wants. Now that I think about it though I might only need this for the policy not for the agent?

>And can't you do everything with torch? (removing the need of numpy)

with the approach I've landed on I certainly can.

>  I know you are trying to have an efficient implementation by using n_envs and and the candidates in parallel,
but I think it would be clearer (and less prone to errors) to use evaluate_policy as in the CEM implementation.

Yes after going through the trouble of making a parallel version I sort of a agree, it won't take long to make the simple version and profile it to see how much performance that leaves on the table.

sgillen:
> That'd be nice. A simple table of the final agent performances side-by-side (with the variances included) would be the best case. We can then determine if it is close enough or something needs tweaking. Again, repeating experiments on the environments original paper used is enough (or on a subset of them).

Ok, I'll make the changes @araffin already suggested, and then rerun my own along with the reference on the v2 Mujoco envs.

sb3_contrib/ars/ars.py
sgillen: It might yes, I generally prefer to keep things in numpy where possible (because I don't need the gradient information, and multiprocessing with torch tensors has gotchas) but there is no good reason to in this context.

sgillen:
Ok, this is finally ready for review, no rush from my end though. Results actually do match the original paper now, I think this can be attributed to:
1. Removing the bias term on the output of the linear policies, which helped the swimmer avoid local minima. 
2. Realizing there was a default max_steps parameter in zoos all_plots.py script (notice in the original plots all had 2e6 time-steps).

I am still running the slower environments, they look like they are matching performance but I will update when they finish.  

I did wind up going with the simpler, essentially single threaded approach using evaluate_policy, it's can't make use of vectorized envs for speedup, but its cleaner, and for the most part one wants to be training on many seeds / params in parallel anyway, so you can still use your whole processor if you want.  


Results_Swimmer-v2
Results_HalfCheetah-v2
Results_Hopper-v2

Original results:
image

araffin:
@sgillen could you have a look at what I did? (I added async evaluation and sync of the normalization, so I can reach 10 000 FPS when training on mujoco =))
I could reproduce your results in the zoo, so what's missing is some polishing (adding docstrings, help is welcomed ;)) and adding some pre-trained models in the zoo (I'll do that).

sb3_contrib/ars/ars.py
```diff
@@ -0,0 +1,289 @@
...
+            policy_base=policy_base,
+            policy_kwargs=policy_kwargs,
+            verbose=verbose,
+            device=device,
+            supported_action_spaces=(gym.spaces.Box, gym.spaces.Discrete),
+            support_multi_env=True,
+            seed=seed,
+        )
+
+        if self.device != th.device("cpu"):
```
sgillen, line 81: If we are forcing the device to CPU should we just remove the device parameter entirely?

sgillen:
Cool! This looks good to me, glad you found a way to make things much faster. I can add some polish/docstrings early next week. 

The one thing that sticks out to me now is the device argument, since the implementation is CPU only now should we just remove that entirely? or do you want to keep it for compatibility? When running on my local machine I see four of the pytests failing right now, two are from TRPO, and two are from ARS not respecting the device argument.

araffin:
>  I can add some polish/docstrings early next week.

perfect =)

> since the implementation is CPU only now should we just remove that entirely? or do you want to keep it for compatibility?

I want to keep it for API consistency and or I plan to make it compatible with GPU at some point (even tough CPU will be the default device).

>  When running on my local machine I see four of the pytests failing right now, two are from TRPO, and two are from ARS not respecting the device argument.

oh, because of the GPU? (I think this is the case for ARS loading at least)
Let me try out, otherwise please open issue with the associated error and tracebacks ;)

sgillen:
Ok, made some minor changes to the doc strings, and all my local tests are passing now too. Let me know if there is anything else I can do.

araffin:
> Ok, made some minor changes to the doc strings, and all my local tests are passing now too. Let me know if there is anything else I can do.

thanks =)
there are some doctrings missing in the ars implementation, for instance for the `do_one_update()`

araffin (approved):
LGTM, thanks =)

sgillen:
> thanks =) there are some doctrings missing in the ars implementation, for instance for the `do_one_update()`

Thanks for taking care of that, I was going from the TQC implementation in terms of what needs a doc string or not.

PR #63, opened 2022-03-07
Speedup LSTM forward pass

Walon1998:
If the episode is never restarted, it is not necessary to manually unroll the LSTM to reset the LSTM states. However, one call to the underlying implementation is sufficient. In a small example I was able to achieve a speedup of 3-4 (RTX 3080 Ti).

### DIFF
--- a/sb3_contrib/common/recurrent/policies.py
+++ b/sb3_contrib/common/recurrent/policies.py
@@ -148,6 +148,11 @@ class RecurrentActorCriticPolicy(ActorCriticPolicy):
         features_sequence = features.reshape((n_envs, -1, lstm.input_size)).swapaxes(0, 1)
         episode_starts = episode_starts.reshape((n_envs, -1)).swapaxes(0, 1)
 
+        if th.all(episode_starts == 0.0):
+            hidden_eff, lstm_states_eff = lstm(features_sequence, lstm_states)
+            hidden_eff = th.flatten(hidden_eff.transpose(0, 1), start_dim=0, end_dim=1)
+            return hidden_eff, lstm_states_eff
+
         lstm_output = []
         # Iterate over the sequence
         for features, episode_start in zip_strict(features_sequence, episode_starts):
--- a/sb3_contrib/common/recurrent/policies.py
+++ b/sb3_contrib/common/recurrent/policies.py
@@ -148,10 +148,12 @@ class RecurrentActorCriticPolicy(ActorCriticPolicy):
         features_sequence = features.reshape((n_envs, -1, lstm.input_size)).swapaxes(0, 1)
         episode_starts = episode_starts.reshape((n_envs, -1)).swapaxes(0, 1)
 
+        # If we don't have to reset the state in the middle of a sequence
+        # we can avoid the for loop, which speeds up things
         if th.all(episode_starts == 0.0):
-            hidden_eff, lstm_states_eff = lstm(features_sequence, lstm_states)
-            hidden_eff = th.flatten(hidden_eff.transpose(0, 1), start_dim=0, end_dim=1)
-            return hidden_eff, lstm_states_eff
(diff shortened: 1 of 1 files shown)


araffin:
Hello,
thanks for the PR =)
will try to review tomorrow, should speedup things for sure, I will just check that the results are the same.

Miffyli:
Looking good to me but I let araffin do the proper check. Meanwhile please check the steps in [PR template](https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/blob/master/.github/PULL_REQUEST_TEMPLATE.md) (at least the changelog and contributors one :) )

araffin:
> Looking good to me but I let araffin do the proper check. Meanwhile please check the steps in [PR template](https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/blob/master/.github/PULL_REQUEST_TEMPLATE.md) (at least the changelog and contributors one :) )

no need here to fill full PR template, as it is a special case: PR for an opened PR ;)

araffin:
I tested it using 
```python
            lstm_output1, lstm_states1 = lstm(features_sequence, lstm_states)
            lstm_output1 = th.flatten(lstm_output1.transpose(0, 1), start_dim=0, end_dim=1)

            lstm_output2 = []
            # Iterate over the sequence
            for features, episode_start in zip_strict(features_sequence, episode_starts):
                hidden, lstm_states = lstm(
                    features.unsqueeze(dim=0),
                    (
                        (1.0 - episode_start).view(1, n_envs, 1) * lstm_states[0],
                        (1.0 - episode_start).view(1, n_envs, 1) * lstm_states[1],
                    ),
                )
                lstm_output2 += [hidden]
            # Sequence to batch
            lstm_output2 = th.flatten(th.cat(lstm_output2).transpose(0, 1), start_dim=0, end_dim=1)

            assert th.allclose(lstm_output1, lstm_output2)
```
and the test passes, but the code somehow changes the random generator state... (I get slightly different results with your PR, but it runs much faster).

Walon1998:
> I tested it using
> 
> ```python
> lstm_output1, lstm_states1 = lstm(features_sequence, lstm_states)
> lstm_output1 = th.flatten(lstm_output1.transpose(0, 1), start_dim=0, end_dim=1)
> 
> lstm_output2 = []
> # Iterate over the sequence
> for features, episode_start in zip_strict(features_sequence, episode_starts):
>     hidden, lstm_states = lstm(
>         features.unsqueeze(dim=0),
>         (
>             (1.0 - episode_start).view(1, n_envs, 1) * lstm_states[0],
>             (1.0 - episode_start).view(1, n_envs, 1) * lstm_states[1],
>         ),
>     )
>     lstm_output2 += [hidden]
> # Sequence to batch
> lstm_output2 = th.flatten(th.cat(lstm_output2).transpose(0, 1), start_dim=0, end_dim=1)
> 
> assert th.allclose(lstm_output1, lstm_output2)
> ```
> 
> and the test passes, but the code somehow changes the random generator state... (I get slightly different results with your PR, but it runs much faster).
> 
> EDIT: the results seem to change also with the number of threads used... not an used with that PR then

Also see https://pytorch.org/docs/stable/generated/torch.nn.LSTM.html: 
grafik

araffin:
thanks but my tests were cpu only... and things are reproducible, just not the same results depending on code used and number of threads

vwxyzjn:
This makes intuitive sense to me. I'd be also interested in seeing the performance comparison.

araffin:
After further investigation, the difference in results is probably due to numerical imprecision (the test actually fails after a while but still passes with `assert th.allclose(lstm_output1, lstm_output2, rtol=1e-05, atol=1e-07)` instead of `assert th.allclose(lstm_output1, lstm_output2, rtol=1e-05, atol=1e-08)`)

PR #120, opened 2022-11-28
Fix `sb3_contrib/common/utils.py` type hint

qgallouedec:
## Context

## Types of changes

## Checklist:

- [ ] I've read the [CONTRIBUTION](https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/blob/master/CONTRIBUTING.md) guide (**required**)

### DIFF
--- a/sb3_contrib/common/utils.py
+++ b/sb3_contrib/common/utils.py
@@ -114,16 +114,12 @@ def conjugate_gradient_solver(
 
     p = residual.clone()
 
-    for i in range(max_iter):
-        # A @ p (matrix vector multiplication)
-        A_dot_p = matrix_vector_dot_fn(p)
+    # A @ p (matrix vector multiplication)
+    A_dot_p = matrix_vector_dot_fn(p)
+    alpha = residual_squared_norm / p.dot(A_dot_p)
+    x += alpha * p
 
-        alpha = residual_squared_norm / p.dot(A_dot_p)
-        x += alpha * p
-
-        if i == max_iter - 1:
-            return x
-
+    for _ in range(max_iter - 1):
         residual -= alpha * A_dot_p
         new_residual_squared_norm = th.matmul(residual, residual)
 
@@ -133,6 +129,10 @@ def conjugate_gradient_solver(
         beta = new_residual_squared_norm / residual_squared_norm
         residual_squared_norm = new_residual_squared_norm
         p = residual + beta * p
+        A_dot_p = matrix_vector_dot_fn(p)
+        alpha = residual_squared_norm / p.dot(A_dot_p)
+        x += alpha * p
+    return x
 
 
 def flat_grad(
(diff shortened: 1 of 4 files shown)


sb3_contrib/common/utils.py
```diff
@@ -114,16 +114,12 @@ def conjugate_gradient_solver(
...
     p = residual.clone()
 
-    for i in range(max_iter):
-        # A @ p (matrix vector multiplication)
-        A_dot_p = matrix_vector_dot_fn(p)
-
-        alpha = residual_squared_norm / p.dot(A_dot_p)
-        x += alpha * p
-
-        if i == max_iter - 1:
```
araffin, line 124: mypy didn't like early return?

I need to check that one, but I think it is ok sometimes to ignore mypy warnings/errors if it is for readability.

qgallouedec: mypy think the function could never return

araffin: is there a case where it would never return?
if `max_iter=0`, the loop doesn't run, if `max_iter=1`, then it returns automatically, I guess it is a false positive?

qgallouedec: >  it is ok sometimes to ignore mypy warnings/errors if it is for readability.

Agree. But here, I've find the new version more readable.

araffin: oh, I see, maybe `assert max_iter > 0` would be a better fix?

qgallouedec: > if max_iter=0, the loop doesn't run

In that cas it is supposed to return None then. Is it the desired behaviour?

qgallouedec:
```python
    assert max_iter > 0
    for i in range(max_iter):
        if i == max_iter - 1:
            return x
```

mypy:
```shell
error: Missing return statement  [return]
```

araffin: mmh, probably too advanced for mypy, I guess we can ignore it then

araffin: > In that case it would return None then. Is it the desired behaviour?

calling the CG solver with `max_iter=0` doesn't make much sense, but adding `return x` at the end should solve the issue, no? (so not changing the rest, even though the return statement will almost never be used)

qgallouedec: > adding `return x` at the end should solve the issue, no?

It would.

The following two structure would work.

1.
```python
for i in range(max_iter):
    # Do A
    if i == max_iter - 1:
         return 
    # Do B
return 
```

2.
```python
# Do A
for _ in range(max_iter - 1):
    # Do B
    # Do A
return
```

I find the second one more readable, but if you prefer to limit the number of changes, we can keep the first one.

araffin (commented):
I did the one line change ;)
tell me if you think we need to had a "run" test for that case.

qgallouedec:
> tell me if you think we need to had a "run" test for that case.

I think it's not necessary.

araffin (approved):
LGTM =)

PR #127, opened 2022-12-10
DuelingDQN

qgallouedec:
## TODOs

- [ ] Make learning curves 
- [ ] Complete DuelingDQN short presentation
- [ ] Fill the result tabular

## Context

## Types of changes

## Checklist:

- [ ] I've read the [CONTRIBUTION](https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/blob/master/CONTRIBUTING.md) guide (**required**)

### DIFF
--- a/sb3_contrib/dueling_dqn/dueling_dqn.py
+++ b/sb3_contrib/dueling_dqn/dueling_dqn.py
@@ -1,0 +1,125 @@
+from typing import Any, Dict, Optional, Tuple, Type, TypeVar, Union
+
+import torch as th
+from stable_baselines3.common.buffers import ReplayBuffer
+from stable_baselines3.common.type_aliases import GymEnv, MaybeCallback, Schedule
+from stable_baselines3.dqn.dqn import DQN
+
+from sb3_contrib.dueling_dqn.policies import DuelingDQNPolicy
+
+SelfDuelingDQN = TypeVar("SelfDuelingDQN", bound="DuelingDQN")
+
+
+class DuelingDQN(DQN):
+    """
+    Dueling Deep Q-Network (Dueling DQN)
+
+    Paper: https://arxiv.org/abs/1511.06581
+
+    :param policy: The policy model to use (MlpPolicy, CnnPolicy, ...)
+    :param env: The environment to learn from (if registered in Gym, can be str)
+    :param learning_rate: The learning rate, it can be a function
+        of the current progress remaining (from 1 to 0)
+    :param buffer_size: size of the replay buffer
+    :param learning_starts: how many steps of the model to collect transitions for before learning starts
+    :param batch_size: Minibatch size for each gradient update
+    :param tau: the soft update coefficient ("Polyak update", between 0 and 1) default 1 for hard update
+    :param gamma: the discount factor
+    :param train_freq: Update the model every ``train_freq`` steps. Alternatively pass a tuple of frequency and unit
+        like ``(5, "step")`` or ``(2, "episode")``.
+    :param gradient_steps: How many gradient steps to do after each rollout (see ``train_freq``)
+        Set to ``-1`` means to do as many gradient steps as steps done in the environment
+        during the rollout.
+    :param replay_buffer_class: Replay buffer class to use (for instance ``HerReplayBuffer``).
+        If ``None``, it will be automatically selected.
+    :param replay_buffer_kwargs: Keyword arguments to pass to the replay buffer on creation.
+    :param optimize_memory_usage: Enable a memory efficient variant of the replay buffer
+        at a cost of more complexity.
+        See https://github.com/DLR-RM/stable-baselines3/issues/37#issuecomment-637501195
+    :param target_update_interval: update the target network every ``target_update_interval``
+        environment steps.
+    :param exploration_fraction: fraction of entire training period over which the exploration rate is reduced
+    :param exploration_initial_eps: initial value of random action probability
+    :param exploration_final_eps: final value of random action probability
+    :param max_grad_norm: The maximum value for the gradient clipping
+    :param tensorboard_log: the log location for tensorboard (if None, no logging)
+    :param policy_kwargs: additional arguments to be passed to the policy on creation
+    :param verbose: Verbosity level: 0 for no output, 1 for info messages (such as device or wrappers used), 2 for
+        debug messages
+    :param seed: Seed for the pseudo random generators
+    :param device: Device (cpu, cuda, ...) on which the code should be run.
+        Setting it to auto, the code will be run on the GPU if possible.
+    :param _init_setup_model: Whether or not to build the network at the creation of the instance
+    """
+
+    def __init__(
+        self,
+        policy: Union[str, Type[DuelingDQNPolicy]],
+        env: Union[GymEnv, str],
+        learning_rate: Union[float, Schedule] = 0.0001,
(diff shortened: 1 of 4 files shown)


araffin:
Just so you know, I had in mind to implement rainbow: https://github.com/DLR-RM/stable-baselines3/issues/622
and make double dqn / dueling dqn special case of it.
But so far I had no time, so I think it's good to have reference implementation here =)

btw, double dqn is implemented here: https://github.com/osigaud/stable-baselines3/blob/feat/double-dqn/stable_baselines3/dqn/dqn.py#L170-L176

Related: https://github.com/DLR-RM/stable-baselines3/issues/487

qgallouedec:
> Just so you know, I had in mind to implement rainbow: https://github.com/DLR-RM/stable-baselines3/issues/622
and make double dqn / dueling dqn special case of it.

Agree, it would be much better

> btw, double dqn is implemented here: https://github.com/osigaud/stable-baselines3/blob/feat/double-dqn/stable_baselines3/dqn/dqn.py#L170-L176

Thanks for the ref!

qgallouedec:
@araffin I want your opinion on this:
Wouldn't it make more sense to modify DQN to make DuelingDQN a special case of DQN? Rather than rewriting DQN, as I am doing on this PR.
Eventually, we would have the same structure as DDPG/TD3.
Moreover, the code written would be more easily reusable for Rainbow.

sb3_contrib/dueling_dqn/dueling_dqn.py
```diff
@@ -0,0 +1,132 @@
...
+    :param _init_setup_model: Whether or not to build the network at the creation of the instance
+    """
+
+    policy_aliases: Dict[str, Type[BasePolicy]] = {
+        "MlpPolicy": MlpPolicy,
+        "CnnPolicy": CnnPolicy,
+        "MultiInputPolicy": MultiInputPolicy,
+    }
+
+    def __init__(
```
araffin, line 62: if __init__ is the same as DQN, I guess you can drop it.

araffin:
> e to modify DQN to make DuelingDQN a special case of DQN?

you mean DQN a special case of DuelingDQN?
how would that look like?

> Rather than rewriting DQN, as I am doing on this PR.

you are mainly changing the policy, no?

sb3_contrib/dueling_dqn/dueling_dqn.py
qgallouedec: The only (but necessary) diff is the policy type hint.

qgallouedec:
> you are mainly changing the policy, no?

I'm changing _only_ the policy.

qgallouedec:
> you mean DQN a special case of DuelingDQN?

Yes, it makes more sense in that order. But this would require putting DuelingDQN in the main repo.
That said, the cleanest way for future implementation would be that DuelingDQN is a special case of RAINBOW (in which all tricks except dueling are disabled), and DQN is another special case of RAINBOW (in which all tricks are disabled).

> how would that look like?

I'll make a draft PR in SB3

sb3_contrib/dueling_dqn/dueling_dqn.py
araffin: I see...

araffin:
> That said, the cleanest way for future implementation would be that DuelingDQN is a special case of RAINBOW 

yes, so for now, I would keep it as is. My goal with Vanilla DQN in SB3 was to keep it as simple as possible to be a reference when learning or implementing other things.

qgallouedec:
On second thought, it might be more practical to work in reverse. I mean, instead of implementing RAINBOW all at once, I was thinking of implementing the features one by one upon DQN, disabling them by default. And at the renaming DQN to RAINBOW, changing the defaults. Roughly we would have:

Currently:

```python
class DQN:
    def __init__(self, policy, env, ...): ...
```

We implement dueling:

```python
class DQN:
    def __init__(self, policy, env, ..., dueling=False): ...

class DuelingDQN(DQN):
    def __init__(self, policy, env, ...):
        super().__init__(policy, env, ..., dueling=True)
```

Then we implement PER

 ```python
class DQN:
    def __init__(self, policy, env, ..., dueling=False, per=False): ...

class DuelingDQN(DQN):
    def __init__(self, policy, env, ..., per=False):
        super().__init__(policy, env, ..., dueling=True, per=per)

class PERDQN(DQN):
    def __init__(self, policy, env, ..., dueling=False):
        super().__init__(policy, env, ..., dueling=dueling, per=True)
```

And so on.

Once all the features have been implemented, we can rename DQN to RAINBOW, changing only the default values, and add a DQN class that inherits from RAINBOW:

 ```python
class RAINBOW:
    def __init__(self, policy, env, ..., dueling=True, per=True):  ...

class DuelingDQN(RAINBOW):
    def __init__(self, policy, env, ..., per=False):
        super().__init__(policy, env, ..., dueling=True, per=per)

class PERDQN(RAINBOW):
    def __init__(self, policy, env, ..., dueling=False):
        super().__init__(policy, env, ..., dueling=dueling, per=True)

class DQN(RAINBOW):
    def __init__(self, policy, env, ...):
        super().__init__(policy, env, ..., dueling=False, per=False)
```
The main constraint is that when renaming DQN -> RAINBOW, and then creating a DQN class that inherits from RAINBOW, we introduce a breaking change, since arguments from DQN are removed (per, dueling, etc.). They would have to be deprecated first.

araffin:
>  I mean, instead of implementing RAINBOW all at once, I was thinking of implementing the features one by one upon DQN, disabling them by default.

why not implement all at once?

Compared to DDPG vs TD3, obtaining DQN from Rainbow is not trivial (correct me if I'm wrong). As DQN is a key algorithm in RL, I would really like to keep it simple so people can study it (even if it means duplicated code when implementing DQN extensions).
And I would see two "family" of algorithms: DQN vs RAINBOW (with more or less extensions), so `PERDQN`, `DuelingDQN`, ... would all derive from `RAINBOW` but not `DQN`.

What you propose is what I had in mind (so Dueling special case of RAINBOW) except for DQN.

qgallouedec:
> why not implement all at once?

We could.
I don't know exactly what changes will be needed in the library to implement RAINBOW. But let's take PER as an example. I think it would be best to implement it as HER, i.e. as a subclass of ReplayBuffer. To make PER compatible with all off-policy algorithms, we will have to find a way to update the weights after the TD error calculation. Probably we will have to add in all algorithms a call to the method `self.replay_buffer.update_weight(td_errors)` or (or `self.replay_buffer.update_weight(idxs, td_errors)`, depending on how we go about it), and thus add this method to the parent class `ReplayBuffer`.
All in all, it shouldn't be complicated but it concerns a lot of files, so I think it would be much easier to work on it on a dedicated PR.

I don't know if the same reasoning applies to other rainbow features.

> Compared to DDPG vs TD3, obtaining DQN from Rainbow is not trivial (correct me if I'm wrong).

Right now I imagine that it should not be complicated, but I may be wrong, I would have to look into it in detail. 

> As DQN is a key algorithm in RL, I would really like to keep it simple so people can study it (even if it means duplicated code when implementing DQN extensions). And I would see two "family" of algorithms: DQN vs RAINBOW (with more or less extensions), so `PERDQN`, `DuelingDQN`, ... would all derive from `RAINBOW` but not `DQN`.

So DQN would be a special case for which we would keep a readable implementation by not deriving it from RAINBOW, right? It seems legitimate.

PR #243, opened 2024-05-04
Implemented CrossQ

### LINKED ISSUE #238: [Feature Request] Implement CrossQ
### 🚀 Feature

I would like to implement CrossQ (https://openreview.net/pdf?id=PczQtTsTIX) in SB3, as also suggested by @araffin (https://github.com/araffin/sbx/pull/36#issuecomment-2027392759), 

### Motivation

CrossQ is one of the current state-of-the-art deep reinforcement learning methods in terms of sample efficiency and substantionally more computationally efficefficient than the previous state-of-the-art (e.g. DroQ or REDQ), as it uses a low update-to-data ratio of 1. It is the first successful application of batch normalization within deep reinforcement learning, which is at the heart of it's efficiency. I think a PyTorch based reference implementation in SB3 would be very valuable for the research community.

### Pitch

As one of the first authors on the paper, I want to contribute a PyTorch based reference implementation of CrossQ to SB3, since the paper's implementation is in JAX.

### Alternatives

_No response_

### Additional context

_No response_

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/issues) in the repo
- [X] If I'm requesting a new feature, I have proposed alternatives

danielpalen:
This PR implements CrossQ (https://openreview.net/pdf?id=PczQtTsTIX), a novel off-policy deep RL algorithm that carefully uses batch normalisation and removes target networks to achieve state-of-the-art sample efficiency at a much lower computational complexity, as it does not require large update-to-data-ratios.

## Description

This implementation is a PyTorch implementation based on the original JAX implementation (https://github.com/adityab/CrossQ).
The following plot shows that the performance matches the performance reported in the original paper, as well as the performance of the open source SBX implementation provided by the authors (evaluated on 10 seeds). 

sbx_reproduce

Open RL benchmark report: https://wandb.ai/openrlbenchmark/sb3-contrib/reports/SB3-Contrib-CrossQ--Vmlldzo4NTE2MTEx

## Context

closes #238

## Types of changes

- [x] Documentation (update in the documentation)

## Checklist:

- [x] I have included baseline results (**required** for new training algorithms or training-related features).


danielpalen:
@araffin in my initial PR it seams like one code style check was failing, sorry about that. I fixed it and it passes on my machine now. I hope it will go through now :)

docs/modules/crossq.rst
```diff
@@ -0,0 +1,99 @@
...
+CrossQ is a simple and efficient algorithm that uses batch normalization to improve the sample efficiency of off-policy deep reinforcement learning algorithms.
+It is based on the idea of carefully introducing batch normalization layers in the critic network and dropping target networks.
+This yield a simpler and more sample-efficient algorithm without requiring high update-to-data ratios.
+
+.. rubric:: Available Policies
+
+.. autosummary::
+    :nosignatures:
+
+    MlpPolicy
```
araffin, line 22: Could you add at least the multi input policy? (so we can try it in combination with HER)
Only the feature extractor should be changed normally.

And what do you think about adding `CnnPolicy`?

docs/modules/crossq.rst
```diff
@@ -0,0 +1,99 @@
...
+  import numpy as np
+
+  from sb3_contrib import CrossQ
+
+  model = CrossQ("MlpPolicy", "Walker2d-v4")
+  model.learn(total_timesteps=1_000_000)
+  model.save("crossq_walker")
+
+
+Results
```
araffin, line 66: Please open a PR in the RL Zoo that adds the hyperparameters used.
So people should be able to reproduce the plots using RL Zoo cli.

sb3_contrib/crossq/policies.py
```diff
@@ -0,0 +1,448 @@
...
+
+        action_dim = get_action_dim(self.action_space)
+        latent_pi_net = create_mlp(features_dim, -1, net_arch, activation_fn)
+
+        if batch_norm:
+            # If batchnorm, then we want to add torch.nn.Batch_Norm layers before every linear layer
+            net: List[Union[nn.Module, BatchRenorm1d]] = []
+            for layer in latent_pi_net:
+                if isinstance(layer, nn.Linear):
+                    net.append(BatchRenorm1d(layer.in_features, eps=0.001, momentum=0.01))
```
araffin, line 93: Please allow to pass the batch norm momentum as parameter

sb3_contrib/crossq/policies.py
```diff
@@ -0,0 +1,448 @@
...
+        self.log_std_init = log_std_init
+        self.use_expln = use_expln
+        self.full_std = full_std
+        self.clip_mean = clip_mean
+
+        action_dim = get_action_dim(self.action_space)
+        latent_pi_net = create_mlp(features_dim, -1, net_arch, activation_fn)
+
+        if batch_norm:
+            # If batchnorm, then we want to add torch.nn.Batch_Norm layers before every linear layer
```
araffin, line 89: What do you think about updating `create_mlp` to allow to pass normalization layer/dropout?

Similar to what is done in https://github.com/DLR-RM/stable-baselines3/pull/1036 and proposed in https://github.com/DLR-RM/stable-baselines3/issues/1069

sb3_contrib/crossq/policies.py
```diff
@@ -0,0 +1,448 @@
...
+            features_extractor_class,
+            features_extractor_kwargs,
+            optimizer_class=optimizer_class,
+            optimizer_kwargs=optimizer_kwargs,
+            squash_output=True,
+            normalize_images=normalize_images,
+        )
+
+        if net_arch is None:
+            net_arch = {"pi": [256, 256], "qf": [2048, 2048]}
```
araffin, line 330: Please use the same defaults as in sbx: https://github.com/araffin/sbx/blob/42caa65fcbf4c74bd007d995baf0f4bdcdf7f42a/sbx/crossq/policies.py#L195-L198

The original hyperparameters can be set in the RL Zoo hyperparameter file.

sb3_contrib/crossq/crossq.py
```diff
@@ -0,0 +1,330 @@
...
+            # Optimize entropy coefficient, also called
+            # entropy temperature or alpha in the paper
+            if ent_coef_loss is not None and self.ent_coef_optimizer is not None:
+                self.ent_coef_optimizer.zero_grad()
+                ent_coef_loss.backward()
+                self.ent_coef_optimizer.step()
+
+            with th.no_grad():
+                # Select action according to policy
+                self.actor.set_training_mode(False)
```
araffin, line 231: is that needed? `self.actor.set_training_mode(False)` is already set above?
or you meant `self.actor.set_training_mode(True)`?

sb3_contrib/crossq/crossq.py
```diff
@@ -0,0 +1,330 @@
...
+            # 2. From a computational perspective a single forward pass is simply more efficient than
+            #    two sequential forward passes.
+            all_obs = th.cat([replay_data.observations, replay_data.next_observations], dim=0)
+            all_acts = th.cat([replay_data.actions, next_actions], dim=0)
+
+            self.critic.set_training_mode(True)
+            all_q_values = th.cat(self.critic(all_obs, all_acts), dim=1)
+            self.critic.set_training_mode(False)
+            current_q_values, next_q_values = th.split(all_q_values, batch_size, dim=0)
+            current_q_values = th.split(current_q_values, [1, 1], dim=1)  # type: ignore
```
araffin, line 257: I guess what you want here is to reshape the tensor? (instead of split)

sb3_contrib/crossq/crossq.py
```diff
@@ -0,0 +1,330 @@
...
+            assert isinstance(critic_loss, th.Tensor)  # for type checker
+            critic_losses.append(critic_loss.item())  # type: ignore[union-attr]
+
+            # Optimize the critic
+            self.critic.optimizer.zero_grad()
+            critic_loss.backward()
+            self.critic.optimizer.step()
+
+            # Compute actor loss
+            self.critic.set_training_mode(False)
```
araffin, line 280: Not needed but maybe for later, we should probably deactivate only the batchnorm? (for instance if dropout is used, we want it to be active there)

sb3_contrib/crossq/crossq.py
```diff
@@ -0,0 +1,330 @@
...
+            q_values_pi = th.cat(self.critic(replay_data.observations, actions_pi), dim=1)
+            self.critic.set_training_mode(False)
+
+            min_qf_pi, _ = th.min(q_values_pi, dim=1, keepdim=True)
```
araffin, line 284: no policy delay for now?
We should probably mention it in the docstring of the class + in the doc

araffin:
Thanks a lot for the implementation =)

I'll try later in the week, but how is it in term of runtime? (SAC vs CrossQ in PyTorch)

danielpalen:
No worries :)

I just pushed most things you requested. I'll add some more specific responses directly to the questions above.

docs/modules/crossq.rst
danielpalen: This is a good point. I looked into it and have not yet added it. If I am not mistaken this would also require some changes to the CrossQ train() function. Since, now concatenating and splitting the batches would also require some control flow based on the used policy.
For simplicity sake (for now) and since I did not have time to try and evaluate the multi input policy I did not add that yet.

docs/modules/crossq.rst
danielpalen: I will do this very soon

sb3_contrib/crossq/policies.py
danielpalen: I think this would make sense, because the way I implemented it right now is really that nice.

sb3_contrib/crossq/crossq.py
danielpalen: I added more mode calls than needed. The reason was, that I wanted to be very specific which more needs to be used where. I think using the wrong BN mode is one of the big gotchas and sources of error when implementing CrossQ. Since this here should be a PyTorch reference to aid others when they want to implement it by themselves I think it is helpful to make the mode very specific to clear up possible confusion.

sb3_contrib/crossq/crossq.py
danielpalen: Same thinking as above. But you are right, if we want to additionally use dropout we should adapt this. Maybe we can just have a set_bn_training_mode function. However, that would be very specific for our use case.

sb3_contrib/crossq/crossq.py
danielpalen: Good catch. I forgot adding it, since in our experiments it's not needed for sample efficiency. I added it now, since the policy delay increases the compute efficiency.

docs/modules/crossq.rst
danielpalen: @araffin I have just opened a RP in RL Zoo https://github.com/DLR-RM/rl-baselines3-zoo/pull/453

araffin:
I'm suspecting something is wrong with the current implementation (I'm currently investigating if it is my changes or not).
My setting:
```yaml
BipedalWalker-v3:
  n_timesteps: !!float 2e5
  policy: 'MlpPolicy'
  buffer_size: 300000
  gamma: 0.98
  learning_starts: 10000
  policy_kwargs: "dict(net_arch=dict(pi=[256, 256], qf=[1024, 1024]))"
```

With the RL Zoo cli for both SBX and SB3 (see SBX readme to have support)
```
python train.py --algo crossq --env BipedalWalker-v3 -P --verbose 0 -param n_envs:30 gradient_steps:30 -n 200000
```
I'm getting much better results with SBX...
I hope it is not the Adam parameters.

danielpalen:
Did you figure out what the issue is? I was at ICRA until last week so I didn't have time but if you didn't find it yet I can also have a look.

Before I pushed my last commit I benchmarked it and there the results looked as expected.

araffin:
> Did you figure out what the issue is? I was at ICRA until last week so I didn't have time but if you didn't find it yet I can also have a look.

not yet, I was on holidays...

> Before I pushed my last commit I benchmarked it and there the results looked as expected.

I mostly observed the discrepancy on the provided env `BipedalWalker-v3` and it seems to be there before my changes.
For the other, I didn't have time yet to launch full benchmark.

araffin:
nevermind, I did some more systematic tests and I couldn't see any significant different, the implementation looks good =)

Report: https://wandb.ai/openrlbenchmark/sb3-contrib/reports/SB3-Contrib-CrossQ--Vmlldzo4NTE2MTEx

danielpalen:
Awesome, let me know if you need anything else :)

araffin:
>  let me know if you need anything else :)

sure, I need to find some time to go over it and maybe polish things here and there.
I will probably delay the custom layer integrated in SB3 to later.

sb3_contrib/common/torch_layers.py
```diff
@@ -0,0 +1,117 @@
+import torch
+
+__all__ = ["BatchRenorm1d", "BatchRenorm"]
+
+
+class BatchRenorm(torch.jit.ScriptModule):
```
araffin, line 6: btw, why did you choose to use a `ScriptModule`, is it faster? if so, how much?

sb3_contrib/crossq/crossq.py
```diff
@@ -0,0 +1,336 @@
...
+            # We need to sample because `log_std` may have changed between two gradient steps
+            if self.use_sde:
+                self.actor.reset_noise()
+
+            # Note: in the following lines we always need to make sure to set train/eval modes
+            # of actor and critic carefully. This is because of the BatchNorm layers in the networks
+            # which behave differently in train and eval modes.
+
+            self.actor.set_training_mode(True)
+            actions_pi, log_prob = self.actor.action_log_prob(replay_data.observations)
```
araffin, line 208: minor but still a difference with original implementation: the batch norm stats for the actor are updated at every training step?
and the ent coef is updated at every training step too?
(in other word, the delay parameter is not taken into account?)

sb3_contrib/common/torch_layers.py
danielpalen: tbh, I don't remember, why I used it. And I just found that it seems to have been deprecated in a while anyway, so we should probably just use nn.Module instead (https://discuss.pytorch.org/t/whats-the-difference-between-torch-nn-module-and-torch-jit-scriptmodule/64480)

sb3_contrib/crossq/crossq.py
danielpalen: good catch, that's true. I guess the easiest fix for this would be to set `self.actor.set_training_mode(update_actor_and_temperature)`.
What do you think?

araffin: i reorganize things and now it should be as in the original implementation

araffin:
I simplified the network creation (need https://github.com/DLR-RM/stable-baselines3/pull/1975 to be merged with master), added the updated beta for Adam (it had an impact on my small experiments with Pendulum) and fixed a wrong default value for BN momentum that I introduced.

araffin:
Sorry for the delay, I finally had some time to run more experiments and it looks good =): https://wandb.ai/openrlbenchmark/sb3-contrib/reports/SB3-Contrib-CrossQ--Vmlldzo4NTE2MTEx

Now, we need to update the doc and it should be ready to merge!

araffin (approved):
LGTM, thanks again =)

